Issue #8, opened 2020-03-22
Add Weblinks


namanmodani:
Append logos that link to our **Twitter** (already present), **Instagram** and **Facebook** handle in order to make it easier for the social team to propagate news and infographics in the general public.

selva-ts:
It seems Insta not allowing to share from another website. Also, we can include LinkedIn.

caffeinatedHuman:
Adding `LinkedIn` doesn't make sense to me.

jeremyphilemon:
We're not actively working on supporting social icons on the frontpage at the moment.

PR #11, opened 2020-03-22
Tweet Intent

jeremyphilemon:
Added a tweet button, need to configure the tweet intent.

<p align="center">
Screen Shot 2020-03-22 at 9 28 06 PM
</p>

Referencing #8


jeremyphilemon:
Too bad :/

PR #14, opened 2020-03-23
Tweet button

selva-ts:
The tweet button (widget) added. But we need to create a new component for the Button and summary calculation should be moved to utils as it is repeated again in other component.

### DIFF
--- a/src/components/home.js
+++ b/src/components/home.js
@@ -17,6 +17,7 @@ function Home(props) {
   const [lastUpdated, setLastUpdated] = useState('');
   const [timeseries, setTimeseries] = useState([]);
   const [deltas, setDeltas] = useState([]);
+  const [summaryString, setSummaryData] = useState('');
 
   useEffect(()=> {
     if (fetched===false) {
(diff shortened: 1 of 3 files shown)


sanair96:
Hey @selva-ts I think there's a few issues with the getSummaryData function.

src/components/home.js
```diff
@@ -32,12 +33,46 @@ function Home(props) {
...
+    let deaths = 0;
+    data.statewise.map((state, index) => {
+      if (index !== 0) {
+        confirmed += parseInt(state.confirmed);
+        recoveries += parseInt(state.recovered);
+        deaths += parseInt(state.deaths);
+      }
+    });
+
+    const todayDate = data.key_values[0].lastupdatedtime.split(',')[0];
```
sanair96, line 56: You will get the last updated at time which may not necessarily be today.

src/components/home.js
```diff
@@ -32,12 +33,46 @@ function Home(props) {
...
+    const todayData = data.cases_time_series.filter((x) => x.date.trim() === todayDate)[0];
```
sanair96, line 57: Same here since you use that data, you will get last updated at's data and not today's data.

src/components/home.js
```diff
@@ -32,12 +33,46 @@ function Home(props) {
...
+    const dailyConfirmed = todayData ? todayData.dailyconfirmed: '0';
```
sanair96, line 58: lastupdated at's confirmed cases

src/components/home.js
```diff
@@ -32,12 +33,46 @@ function Home(props) {
...
+
+    const dateTimeString = new Date().toLocaleString('en-IN', {timeZone: 'Asia/Kolkata'}).slice(0, -3);
```
sanair96, line 60: today's date

src/components/home.js
```diff
@@ -32,12 +33,46 @@ function Home(props) {
...
+
+    Number of cases reported today: ${dailyConfirmed}
+
+    Follow @covid19indiaorg
+
+    #COVID19India #SocialDistancing
+    More @`;
+
+    setSummaryData(twitterString);
+  };
```
sanair96, line 74: The string would show as of today, there are so and so confirmed, recovered, deaths. But the cases reported today would be the number of cases would be the number of confirmed would be as on the last updated at date

selva-ts: Got it, but there is no documentation available and data doesn't consistent. So I have used this logic. Can you please share any document or link to look into?

sanair96:
Hey
Let's not close this. Instead you could change the date in the tweet to match the last updated date

selva-ts:
Actually we have some other merge issues. I have discussed with @jeremyphilemon and decided to close this and open a new PR.

sanair96:
Okay cool

PR #40, opened 2020-03-23
Fix percentage calculation for states

### LINKED ISSUE #39: Error in percentage calculation for states
As seen in the screenshot below, percentages are not calculated properly and are neither rounded off. Using current data, the result should have been 100 * 29 / 440 = 6.59%, instead of 7%.

Screenshot from 2020-03-23 17-03-38

shuklaayush:
Fixes #39

Screenshot from 2020-03-23 17-19-15


jeremyphilemon:
Neat fix!

Issue #19, opened 2020-03-23
Social - Twitter, fb, insta button + scrolling banner


Reddevilsab:
We(Social team) need a scrolling banner on the top of the page which will contain factoids and Mythbusters. All content will be supplied by us. Please make that functionality and we shall give you the content. Much needed as so much fake news circulates in the media nowadays and we want to get out the facts to our 150,000+ Daily viewers.
Second - we need our social media pages to be pinned onto the website so as to divert traffic from the website to SM pages as well. Mainly have an icon for twitter below the 'Last case seen on'. And insta and fb buttons in the right top corner. 
Please complete both of these on priority. 
Thanks.

dakshshah96:
Duplicate of #8

Reddevilsab:
Does it have a mention of scrolling banner on the top of the page with
factoids?

jeremyphilemon:
@JunaidBabu @sudevschiz Thoughts on the banner?

sudevschiz:
@jeremyphilemon Closed the issue for Social Media icons. We will add them later.
But the scrolling banner with Factoids is necessary now I think. Social team has already created this.

Reusing this issue for that work. 

@JunaidBabu , created an issue for that in API https://github.com/covid19india/api/issues/4

zeffirara:
Scrolling banners are hard to read. A carousel style banner with text that keeps changing after a fixed interval would be better I think.

jeremyphilemon:
@kapiljain-128 your wish has been fulfilled! 😌

PR #42, opened 2020-03-23
Social Media Buttons

selva-ts:
Social Media Buttons (Twitter and FB Share) added. A new component socialmedia created. The current FB share button allows sharing the URL. But we need to implement the FB Quote with the pre-populated data.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,9 @@ module.exports = {
     'react',
   ],
   'rules': {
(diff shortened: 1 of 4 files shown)


selva-ts:
The Tweet and Share button aligned with Telegram button.
image

sanair96:
Solves #19 and #8

JunaidBabu:
Will consider social media buttons later. Right now it's taking up space in the site which we can use for useful things

PR #51, opened 2020-03-24
Fix colorbar scheme

### LINKED ISSUE #50: Colorbar scheme is different from map colorscheme
a

shuklaayush:
Fixes #50. Also changed colorbar scale to represent total confirmed cases instead of percentages.


jeremyphilemon:
Good stuff! Thanks for this!

PR #54, opened 2020-03-24
Fix position of hover display

shuklaayush:
1. Increased width and height of textbox for state names to fix position of hover display
2. Moved colorbar lower

Peek 2020-03-24 14-10

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -560,7 +560,8 @@ table {
     }
     h4 {
       color: #e23028;
-      width: 5rem;
+      width: 8rem;
+      height: 2em;
     }
     h1 {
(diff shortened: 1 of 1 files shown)


shuklaayush:
Fixes this annoying transition. @jeremyphilemon Thoughts?

Peek 2020-03-25 13-54

jeremyphilemon:
@shuklaayush The increase in width seems to disable pointer events on Gujarat, check in mobile. Check #90 for an update on this!

shuklaayush:
Yeah, the text seems to be interfering. #90 is a better approach compared to this hack. Closing.

PR #63, opened 2020-03-24
Link to the updated github page

vamsibedapudi:
Old webpage ponts to "github.com/covid19india", changing it to "github.com/covid19india-react"


sudevschiz:
This is by design pointing to the organization page. Hence not merging.
Thanks for the PR!

PR #71, opened 2020-03-24
Hardcode colour of states with zero cases to white

### LINKED ISSUE #70: Colour code of map
Kinda personal but the color code for the states with 0 cases should be white cause sometime the eastern states looks like they are effected.

shuklaayush:
Zero value of interpolateReds( ) scale is not mapped to complete white. Hardcode zero value to white.
Fixes #70.


jeremyphilemon:
Looks good! 🚀

Issue #82, opened 2020-03-25
Hovering out of states in map doesn't change data for that top left panel.


turbanpride:
Clearly, when we hover over a state in the map, the respective data is shown at the top left panel. But on hovering out of the state doesnt revert back the data in the top left panel to total. 

**To Reproduce**
1. Hover to any state in the map.
2. Check the changes on the top left panel of map component which shows data of hovered state.
3. Hover out of the state and the data at the panel remains same. 
![bug](https://user-images.githubusercontent.com/36756462/77507290-24cc4680-6e8e-11ea-99fa-e63c62b9abea.png)

**Expected behavior**
It should have shown the total data on the left top panel of map component. (refer the screenshot attached).

sangamesh1439:
looking into the issue.

turbanpride:
> looking into the issue.

I have solved the issue. commiting it now

sangamesh1439:
@turbanpride : I have also resolved. https://github.com/covid19india/covid19india-react/pull/83

Sorry for duplicate work .

shuklaayush:
I think resetting to state 1 doesn't make sense. Maybe have a total count by default. Thoughts?

turbanpride:
> I think resetting to state 1 doesn't make sense. Maybe have a total count by default. Thoughts?

Well we already have an object containing total values. We just need to check for that object in mouseout handler and set the state as per that. That's what my solution was. I have shared the handler below

`const handleMouseout = (name) => {
    states.map((state, index) => {
      if (state.state.toLowerCase()==="total") {
        setState(state);
        setIndex(index);
      }
    });`

sangamesh1439:
When app loads for first time, we are setting default state as state 1. API returns state list in decreasing order (More counts will be first state) .On MouserEnter we are changing state and onMouseLeave we should revert it. Thats why I am setting it to state 1. Please run the code and check the user expirence. Its looking good.

arushibhat:
It might not be useful for the user to see the data getting reset to some state he's not intending to look at. As a use case i was thinking users can click state after state to compare the data for their own use case. it'll be difficult for the user to comprehend why the data is getting reset to some unkown state.
Thoughts...

sangamesh1439:
In my opinion, first state has more number of corona counts. Thats why I was thinking to select that on mouseLeave . Let me know what exactly needs to do. I will do changes as required.

arushibhat:
I understand.
[...]

PR #83, opened 2020-03-25
Fixed: On mouseLeave set default state as state 1

sangamesh1439:
Issue Reported here: https://github.com/covid19india/covid19india-react/issues/82


sanair96:
Change on leave to state 0 that'll show three total

jeremyphilemon:
Left a comment at #82 , nice work regardless! Thanks!

PR #84, opened 2020-03-25
Fix 404 on direct links

someshkar:
Switched to `@reach/router` to fix 404 on direct links, only problem now is the navbar's 'clicked' state doesn't sync when you visit a subfolder directly. Can be fixed later though.

P.S. Had the idea to switch to `@reach/router` because [Gatsby](https://www.gatsbyjs.org/) uses it in the background, and it works like a charm.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "semi": true,
+  "singleQuote": true
+}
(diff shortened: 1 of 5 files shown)


someshkar:
Nvm, works now with the `build/404.html` fix.

PR #95, opened 2020-03-25
Added active timeseries graph on homepage

anuragjain:
Added time series graph for active cases. This is an important metric to track and know the progress of this epidemic.
Screen Shot 2020-03-25 at 3 51 23 PM


jeremyphilemon:
@JunaidBabu @sudevschiz Thoughts on adding active? I feel like just because active has a similar trend it needs to be ignored, could lead to visualizations like how fast people moved from active to either recovered or deceased state, confirmed does not convey that. But thoughts?

anuragjain:
This is helpful to assess the current situation. Look at the China graph below which showcases the active patients. Even though total cases are above 80k, the active cases are less than 5k and that number going to zero signifies that the country is free of any infection. In my opinion this is very useful graph, let me know your thoughts.
Screen Shot 2020-03-26 at 3 47 55 PM

jeremyphilemon:
We won't be moving forward with adding "active" graphs at the moment.

PR #113, opened 2020-03-25
Use cursor to slide across charts

shuklaayush:
Earlier, focus could only be captured from the circular markers. Added code for capturing focus from any part in the charts.

Peek 2020-03-25 21-16


JunaidBabu:
@jeremyphilemon this is a very much required feature. Seems to be working well. Please review the code once and merge.

PR #115, opened 2020-03-25
Add Prettier

yogeshvar:


JunaidBabu:
@jeremyphilemon do we need this? if not required close this.

jeremyphilemon:
Right now the codebase is using eslint + google's javascript style guideline. Would appreciate an issue raised first so we could discuss on style guidelines! Closing this for now.

Issue #12, opened 2020-03-22
Support State Deltas


jeremyphilemon:
Like given in the preview:

<p align="center">
<img width="574" alt="Screen Shot 2020-03-22 at 9 41 27 PM" src="https://user-images.githubusercontent.com/17938322/77257511-f8de7480-6c85-11ea-9f45-d81a87254dd1.png">
</p>

maganuk:
@jeremyphilemon since the statewise data is coming directly from the google sheet I think you need to build in a formula for the daily change in the sheet itself. Is the sheet publicly available? I can make the changes and also do the react bit.

jeremyphilemon:
@maganuk so we were planning on using the `time_series_cases` fetched from http://api.covid19india.org/data.json and implement an helper function in the client side to calculate and format it to support statewide deltas meanwhile. If you'd like to add that in the api itself, you can prolly look at https://github.com/covid19india/api

jkandasa:
> so we were planning on using the `time_series_cases` fetched from http://api.covid19india.org/data.json and implement an helper function in the client side to calculate and format it to support statewide deltas meanwhile

@jeremyphilemon  I do not see any state reference in `cases_time_series`. You can not do with this data. Needs to update in API `data.json`

anuragjain:
@jkandasa There is a key named "statewise" in the api [https://api.covid19india.org/data.json](https://api.covid19india.org/data.json) which can be used. However to determine delta, you'll need data for previous day as well. For that an api change might be required.

sumitmckv:
@anuragjain @jeremyphilemon @jkandasa  created PR for api change
https://github.com/covid19india/api/pull/8

PR #97, opened 2020-03-25
Added state wise deltas

sumitmckv:
Fixes #12
Depends on #https://github.com/covid19india/api/pull/8
state_wise_deltas_ui

PR #120, opened 2020-03-25
Support State Deltas #12 

shajahan-dev:
State level delta implemented for table


jeremyphilemon:
Patched by an earlier PR #97. Thanks anyway :)

Issue #76, opened 2020-03-25
Replace FAQs with FAQ


shuklaayush:
Plural is implicit in the acronym FAQ. The heading as well as the links would need to be changed.

Rajanpandey:
Done.
PR: https://github.com/covid19india/covid19india-react/pull/134

jeremyphilemon:
Patched by #85

PR #134, opened 2020-03-25
changed faqs to faq

Rajanpandey:
Done as per issue https://github.com/covid19india/covid19india-react/issues/76


jeremyphilemon:
Thank you, but someone had already created a PR(#85) earlier!

Rajanpandey:
@jeremyphilemon Noted.

The follow contributor has replaced FAQs with FAQ in the frontend part. 
But FAQs is still remaining in the code at two places:
1. getFAQs function
2. setView('FAQs')

If these need to be changed in the code also, then you can merge my branch. Thank you.

Issue #119, opened 2020-03-25
State selected in table should be highlighted in map and subsequent details should be shown.


sscsp489:
when we select or hover on a state in table , automatically that respective state will get highlighted in map and subsequent detail should be shown , and it should also happen vice versa !

![Screenshot from 2020-03-25 22-58-32](https://user-images.githubusercontent.com/36526662/77566840-3fcda380-6eec-11ea-9b63-cf80d39dcc1f.png)

shuklaayush:
Relevant: https://css-tricks.com/svg-map-rollovers/

radhasatam:
Created pull request #137 for this.

PR #137, opened 2020-03-26
Highlight state on map when table state is hovered

radhasatam:
When a state in the table is hovered over (or clicked on mobile), the state details on the map are shown. This is for [Issue 119](https://github.com/covid19india/covid19india-react/issues/119)


shuklaayush:
Nice. Can you add highlighting of the state borders on the map as well?

jeremyphilemon:
Looks good! Merging this for now, feel free to open a PR supporting state borders!

PR #138, opened 2020-03-26
Fix the state-wise confirmed cases count (table)

pranavcode:
Simple change to fix the confirmed count for each state.

|Previous|Update|
|-|-|
|Screen Shot 2020-03-26 at 6 55 30 AM|Screen Shot 2020-03-26 at 6 54 10 AM|


jeremyphilemon:
Hey thanks for the quick fix! Patched it before seeing the PR, but thank you!

PR #140, opened 2020-03-26
Bug Fix - Confirmed and Active cases correctly displayed

unmesh-shukla:
Bug Description -
Number of Active Cases is being displayed instead of Number of Confirmed Cases in each row.


jeremyphilemon:
Hey thanks for the quick fix, patched it before I saw this! But shh don't tell anyone how silly the error was 👀

PR #141, opened 2020-03-26
Added tab in nav for Deep Dive

prtksxna:
Created a new version of the dashboard without the toolbar so
that it looks proper with the UI.

### DIFF
--- a/src/components/navbar.js
+++ b/src/components/navbar.js
@@ -61,6 +61,20 @@ function Navbar(props) {
           </Link>
 
           <Link
+            to="/deepdive"
+            onClick={() => {
+              setView("DeepDive");
+            }}
+          >
+            <span
+              className={`fadeInUp ${view === "Network Map" ? "focused" : ""}`}
+              style={{ animationDelay: "0.3s" }}
+            >
+              Deep dive
+            </span>
+          </Link>
+
+          <Link
             to="/links"
             onClick={() => {
               setView("Helpful Links");
--- a/src/components/navbar.js
+++ b/src/components/navbar.js
@@ -53,7 +53,7 @@ function Navbar(props) {
             }}
           >
             <span
-              className={`fadeInUp ${view === "Network Map" ? "focused" : ""}`}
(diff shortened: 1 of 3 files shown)


src/components/navbar.js
```diff
@@ -60,6 +60,20 @@ function Navbar(props) {
...
           </Link>
 
+          <Link
+            to="/deepdive"
+            onClick={() => {
+              setView("DeepDive");
+            }}
+          >
+            <span
+              className={`fadeInUp ${view === "Network Map" ? "focused" : ""}`}
```
KrishnaPravin: It should be **"DeepDive"** instead of **"Network Map"**.
This line is used to highlight the current selection.

KrishnaPravin (changes requested):
Please make this [tiny change](https://github.com/covid19india/covid19india-react/pull/141/files#r398332967)

src/components/navbar.js
prtksxna: Thanks @KrishnaPravin. Updated!

KrishnaPravin (approved):

prtksxna:
I actually see a few other issues with the existing code. I'll update those too.

prtksxna:
This PR also fixes the issue of the "Clusters" link not being highlighted in the navbar when it is indeed selected.

KrishnaPravin:
@prtksxna There are some minor UI issues when viewing on mobile. Could you please have a look at that as well.

prtksxna:
Hey @KrishnaPravin, I'd be able to look at some of those over the weekend. I'd leave it up to you if you'd like to have this merged till then :shipit:

It would also be a great help if you could raise some issues directly on the repo here - https://github.com/prtksxna/covid19-india-dc

jeremyphilemon:
Hey @prtksxna, thanks for the PR! I know we've discussed about the deep dive section extensively in the dev groups and wanted it to be up, however, at the moment we're trying to cook up some different strategies for this section. I'll be opening a new issue soon regarding more visualizations being shown so that we can have a discussion first and move ahead with the PRs :)

Closing this for now!

PR #144, opened 2020-03-26
Revert to right alignment for numbers in table

### LINKED ISSUE #143: Numerical entries are left aligned in table
#97 switched the alignment of numbers left to accommodate deltas. I think we should revert to right alignment as convention suggests.

shuklaayush:
Fixes #143


jeremyphilemon:
Thanks!

PR #149, opened 2020-03-26
Remember last graph state used by user

bnap00:
Using local storage, we can use the previous status that user used in graphs. 

I generally use same type of graph mode that I like all the time, it is nice to have the setting replicated every time I refresh, it uses the same options in case default is something I don't use. 

The user who does not change those options will feel no difference


jeremyphilemon:
Right now we don't want to write on localStorage, and remembering states does not help much in this case, but thank you!

PR #151, opened 2020-03-26
Add sticky headers on homepage table

### LINKED ISSUE #146: Scrolling header for table
User is unable to know the context of the table data fields when the header goes out of the screen. 

Something like this:

-

**Additional context**
-

bnap00:
Fixes #146

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -463,6 +463,9 @@ table {
       &:hover {
         background: $gray-hover;
       }
+      position: sticky;
+      top: 0px;
+      background-color: white;
       .heading-content {
         display: flex;
         justify-content: space-between;
@@ -528,13 +531,6 @@ table {
     td {
       background: $gray-hover;
     }
-  }
-}
-.stickyTable {
-  thead {
-    position: sticky;
-    top: 0;
-    background-color: white;
   }
 }
 .ChoroplethMap {
--- a/src/App.scss
+++ b/src/App.scss
@@ -530,7 +530,13 @@ table {
     }
   }
 }
-
+.stickyTable {
+  thead {
+    position: sticky;
+    top: 0;
+    background-color: white;
+  }
+}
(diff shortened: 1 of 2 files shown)


jeremyphilemon:
Could you review this again, unable to reproduce it?

src/App.scss
```diff
@@ -530,7 +530,13 @@ table {
     }
   }
 }
-
+.stickyTable {
+  thead {
+    position: sticky;
+    top: 0;
+    background-color: white;
```
jeremyphilemon: Let's not change the background color :)

bnap00: actually it makes the header translucent, and hinders visibility, 
check this out 👇 
Screenshot 2020-03-26 at 12 10 42 PM

bnap00:
works for me 
Screenshot 2020-03-26 at 12 15 10 PM

bnap00:
looks like chrome is not supporting CSS property `position: sticky`, 
but the [artical](https://developers.google.com/web/updates/2016/12/position-sticky) suggests that chrome should support it

satishsingh35:
@bnap00 apply sticky to all 'th', to get 'sticky' to work. 'sticky' does not seem to work on table element 

@jeremyphilemon Can you please add me as contributor to this project?

bnap00:
@satishsingh35 I think that did the trick

shuklaayush:
Just a thought. Should we also change the font size and background of the header row to match the last row (total)?

bnap00:
> Just a thought. Should we also change the font size and background of the header row to match the last row (total)?

the total row is also transparent. 
cannot do anything that is transparent 
Screenshot 2020-03-26 at 3 55 13 PM

jeremyphilemon:
Wasn't able to test this out locally :/

But thank you, closing this because I went with a later submitted patch that worked!

PR #169, opened 2020-03-26
Added support for district wise data below MAP, so that hover on map changes data below

aravindha1895:
Added support for district wise data below MAP, so that hover on map changes data below and also hover over the table 

Hover over app in addition to changing stats on left side, also changes stats below thereby avoiding the need to search for the state in the table and expanding

image


shahrk:
This feature is already added PR #164

jeremyphilemon:
Hey thanks for the PR! The feature seems to have been already implemented. Feel free to open new PRs on accepted issues :)

PR #173, opened 2020-03-26
Update App.scss

unmesh-shukla:
- added border-bottom for better content separation in mobile-view.


jeremyphilemon:
We're using alternative backgrounds for the rows as our strategy for a visual separation and remain to use it until mentioned in the issues, thanks :)

Issue #178, opened 2020-03-26
Use AJAX/XHR to fetch new data every x intervals.


aekanshd:
**Is your feature request related to a problem? Please describe.**
Being a web developer, I assumed the website updates realtime. Having the network tab opened for 15 minutes, there was no XHR call made which forced me to do a manual refresh.

**Describe the solution you'd like**
Have a **Last Local Refresh** counter, and make the website pull new data every "x" intervals.

**Describe alternatives you've considered**
We could do away with the counter, but it is usually helpful to know the last fetch.

**Additional context**
None.

s-naveen:
Adding to that if the tab is kept opened we can send out browser push notifications if there's new cases updated. But I guess we need web socket support in server side to achieve the same.

satishsingh35:
@JunaidBabu What is the process of contribution to this repo? Can you help me add as a collaborator

vinishanto:
I have added a pull request that will do an ajax polling every 5 seconds to update the data live without refresh. https://github.com/covid19india/covid19india-react/pull/202

abhishekkhandait:
#196 closes this.

vinishanto:
#202 covers the home page component that was missed in #196. also included disabling the poll when navigated to other tab

JunaidBabu:
Closed with #196

PR #196, opened 2020-03-26
Use AJAX/XHR to fetch new data every x intervals. #178

shirshandu:
used setInterval to refresh the axios call after every 10mis


abhishekkhandait:
@shirshandu you forgot to clear interval once the user navigates to another tab

PR #195, opened 2020-03-26
Active delta at top summary does not match with the one in table total row

rohit2sharma95:
Fixes #188 

Fixed by showing the data from the statewise array

Issue #188, opened 2020-03-26
Active delta at top summary does not match with the one in table total row.


s-naveen:
**Describe the bug**
Active delta at top summary does not match with the one in table total row

**Screenshots**
https://prnt.sc/rn8jkp

rohit2sharma95:
May I fix this issue @s-naveen? We can show the data from first index of the statewise array that is being used in the total also. See the images:-

![image](https://user-images.githubusercontent.com/26922866/77661619-e2dff500-6fa0-11ea-8ab7-c48217f2bc6a.png)

s-naveen:
@rohit2sharma95 that's what is already used. I guess its a bug in API confirmed delta also has mismatch now.

rohit2sharma95:
@s-naveen Delta in the Level component is being shown from the first index of key_values array:-
![image](https://user-images.githubusercontent.com/26922866/77663160-c2b13580-6fa2-11ea-9707-66a79e630939.png)
Whenever the proper data is at the first index of the statewise array that is not being used in the Level component:-
![image](https://user-images.githubusercontent.com/26922866/77663370-fe4bff80-6fa2-11ea-86d5-7dd5c9d00623.png)

Let me know if it is right?

s-naveen:
I am not sure the one in the key_values json or statewise json first key is correct. Can you check it yourself.

rohit2sharma95:
Okay, @s-naveen. I have checked the data manually. Let me give a pull request immediately.

Rajesh-Narayanappa87:
@rohit2sharma95 : I think we have to either point all 'Confirmed', 'Active', 'Recovered' and 'Deceased' to not to point to 'key_values'.. I think this should be api fix not client side.. Open for suggestions..

rohit2sharma95:
@Rajesh-Narayanappa87 I have fixed the issue by showing the data the same as the total row at the bottom of the table. And if you will see the total is correct there.

Rajesh-Narayanappa87:
@rohit2sharma95 : Did you had a chance to look into recover data.. Since first array has "recovered": 0. , graph will be flat.. Check snapshot below..
This still seems service fix.. If we refer state[0] then recovered data is incorrect. 
If we refer key_values then active calc is incorrect.. 
 
[...]

PR #199, opened 2020-03-26
fix for incorrect active case numbr. Bug 188

Rajesh-Narayanappa87:


Rajesh-Narayanappa87:
#188

rohit2sharma95:
@Rajesh-Narayanappa87 There is a difference between the data in the summary and the data in the total row at the bottom of the table:-

image

image

There is a difference in delta for recoveries. This is fixed in #195 though.

Rajesh-Narayanappa87:
@rohit2sharma95 : Good observation.. But this issue was for only fixing active number from table to summary.. If they want to fix for recoveries, then that should be different issue..

PR #202, opened 2020-03-26
Ajax polling auto refresh

vinishanto:
Simple ajax polling to trigger a state update every 5 seconds(5000ms). added a hook for react interval

### DIFF
--- a/src/components/home.js
+++ b/src/components/home.js
@@ -40,7 +40,7 @@ function Home(props) {
           console.log(err);
         });
   };
-
+  //  Initiate ajax polling to refresh the data every 5 secs
   useInterval(() => {
     getStates();
   }, 5000);
--- a/src/components/home.js
+++ b/src/components/home.js
@@ -9,6 +9,7 @@ import ChoroplethMap from './choropleth';
 import TimeSeries from './timeseries';
 import Minigraph from './minigraph';
 import Banner from './banner';
+import { useInterval} from '../hooks/useInterval';
 
 function Home(props) {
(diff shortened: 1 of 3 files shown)


abhishekkhandait:
refer #196

vinishanto:
That's fine never mind. The issue is still open https://github.com/covid19india/covid19india-react/issues/178. I recommend link this and close the issue

vinishanto:
This PR also covers other components on the home page that was missed in #196. This also includes the feature, that polling will stop when navigated to a different tab and I believe the ideal time for ajax polling is 5 secs.

jeremyphilemon:
Hey this looks good, however, I think polling every 5secs might seem a little too overkill considering the api refreshes every nine-ten minutes. Could you change that value, squash your commits and rebase?

dpnkrpl:
No need to refresh in every 5 seconds. 15 minutes should be enough I guess.

jayantbh:
If the DB updates every 9-10 minutes, then I guess no need for a polling based auto-refresh at all.

Realistically no one has the tab open for 10 minutes. But I reckon the maintainers with access to analytics would know better.

I understand that a PR has been merged for updating every 10 minutes, but I'd like to know your thoughts nonetheless.

shivendra-kumar:
Can't we use Server Sent Events for this instead of Ajax polling. Whenever there will be an update server will just push it to client so need of polling.

jeremyphilemon:
Closing due to inactivity.

PR #211, opened 2020-03-27
Refactored table component, clearInterval on unmount, updated sort logic without state mutations

crup:
# Cleaned table.js
- Removed functions calls from table.js `doSort` and `setInterval`. These function calls should be effects. 
- Fixes existing sort which was mutating the component state instead of setting new one.
- Fixes an effect where `setCount` was being called recursively.
- Cleaned `onHighlightState` method.
- Refactored table rows to array map.


jeremyphilemon:
This looks good! Could you rebase please? :)

crup:
Updated with master.

crup:
Reverted my sort implementation, looks like I forgot to handle `state`'s case in previous PR. So, for my I'm sticking with existing sort logic.

src/components/table.js
```diff
@@ -16,30 +16,28 @@ function Table(props) {
...
+      const affectedStates = props.states.reduce((pv, cv, ci) => {
+        return cv.confirmed > 0 && ci > 0 ? pv + 1 : pv;
+      }, 0);
+      setCount(affectedStates);
     }
-  }, [states.length]);
+  }, [props.states.length]);
 
   useEffect(()=>{
     getDistricts();
```
vramana, line 43: Can you also remove the 1 in the inputs array here? It does not matter at all.

src/components/table.js
```diff
@@ -56,113 +54,101 @@ function Table(props) {
...
+    const interval = setInterval(getDistricts, 600000);
+    return () => clearInterval(interval);
+  }, []);
+
+  const sortByKey = (arr, key, isAscending) => {
+    return arr.concat().sort((s1, s2) => {
+      const v1 = key !== 'state' ? parseInt(s1[key]) : s1[key];
+      const v2 = key !== 'state' ? parseInt(s2[key]) : s2[key];
+      if (isAscending) {
+        return v1 > v2 ? 1 : (v1 == v1) && s1['state'] > s2['state'] ? 1 : -1;
```
vramana: Why not multiply this value by -1 if isAscending is false?  That will remove the else branch.

src/components/table.js
```diff
@@ -56,113 +54,101 @@ function Table(props) {
...
+        return v1 < v2 ? 1 : (v1 == v2) && s1['state'] > s2['state'] ? 1 : -1;
       }
     });
-    {/* console.log(states);*/}
-    states.unshift(totalRow[0]);
   };
 
-  const handleSort = (e, props) => {
+  useEffect(() => {
+    const data = [...props.states].slice(1, props.states.length);
```
vramana:
```js
const data = props.states.slice(1) 
```
This has the same effect.

src/components/table.js
crup: Cool! Fixed it.

src/components/table.js
crup: Fixed it. It was a miss.

src/components/table.js
crup: Done!

src/components/table.js
```diff
@@ -56,117 +54,100 @@ function Table(props) {
...
     });
-    {/* console.log(states);*/}
-    states.unshift(totalRow[0]);
   };
 
-  const handleSort = (e, props) => {
-    const currentsortColumn = e.currentTarget.querySelector('abbr').getAttribute('title').toLowerCase();
-    const isAscending = sortData.sortColumn == currentsortColumn? !sortData.isAscending : sortData.sortColumn === 'state';
+  useEffect(() => {
+    const data = props.states.slice(1, props.states.length);
```
vramana: You don't need to mention props.state.length. It automatically slices till the end if you omit it. 

```
props.states.slice(1)
```
This is good enough

crup: Done!!

jeremyphilemon:
Could you please squash and rebase your commits? :)

crup:
Too much pain to squash and resolve conflicts. Will create a new branch from current master and do the changes.

jeremyphilemon:
@crup I understand, apologies for the backlog and the eventual conflicts! Will keep an eye out for your PR! 

Also create an issue and have yourself assigned so we can keep track of it better!

PR #216, opened 2020-03-27
Fixed time calculation

### LINKED ISSUE #213: Last updated time is incorrect across timezones
**Describe the bug**
Relative last update time should be the same value across all timezones in the world. Users from US report "Last updated time" to be off by about 17 hours.

**To Reproduce**
Use geopeeker to load from a different location and send a screenshot (But surprisingly it didn't work for our website though)

**Expected behavior**
All relative times across regions should be the same

**Screenshots**

photo5181748433852147803
photo6337059855830985054

**Smartphone (please complete the following information):**
 - Device: All
 - OS: All (Assuming based on the response in the group that it was tested on multiple phones)
 - Browser 

**Additional context**
The Timestamps in the Google Sheet is always in IST. The same value will be captured as a string by the API

peteykun:
Fixes https://github.com/covid19india/covid19india-react/issues/213


sudevschiz:
@JunaidBabu Review ^ ?

shuklaayush:
@peteykun Maybe remove the colon from the timezone as in ISO-8601.

jeremyphilemon:
Thank you for the PR! Looks like it's patched now!

PR #221, opened 2020-03-27
Added state wise population column

ganeshpatil0101:


ganeshpatil0101:
Added new column to table ie. population state wise in shorter format.
Please review and merge. And let me know if there any changes.

shuklaayush:
Hi @ganeshpatil0101. The population column overflows to the right on mobile when districts are expanded.

ganeshpatil0101:
@shuklaayush I can see good on my Nokia 6.1 Plus mobile. Can you please share the details for mobile device you are checking. Check the attached screenshot for reference.
I agree that there are no more space to add that column. Can you please suggest where we can add that number rather than adding in column.

Thanks for review.
Screenshot_20200327-164750

s-naveen:
Sadly there's not enough space to add any additional columns. Maybe we can add them only on bigger screens.

shuklaayush:
@ganeshpatil0101 I think @s-naveen is right. We don't have enough space for this information on the table. Maybe we can display it somewhere around the map if a design is accepted. Closing.

image

PR #227, opened 2020-03-27
Change daily trends to stem plot

### LINKED ISSUE #189: Change the Daily trends to bar graph
Since each series (Confirmed, Recovered and Deceased) are separate graphs, the best way to represent this information in DAILY TRENDS is to use bar graph.

In DAILY TRENDS, graphs should be a bar graph

-

**Additional context**
We didn't do this initially because in the first version of the site, we had all three series together on the same plot. In that case, line graph was suitable. Now bar graph is more intuitive.

shuklaayush:
Fixes #189  
- Switched daily trends to stem plots which serve a somewhat similar purpose as bar graphs
- Delete a lot of duplicate code by vectorizing all charting functions

image

@sudevschiz Opinions?

### DIFF
--- a/src/components/timeseries.js
+++ b/src/components/timeseries.js
@@ -55,17 +55,19 @@ function TimeSeries(props) {
     setIndex(timeseries.length-1);
 
     const svg1 = d3.select(graphElement1.current);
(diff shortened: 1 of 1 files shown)


jeremyphilemon:
Looks good! Squash and rebase? :)

Issue #190, opened 2020-03-26
Add more informational regarding Transmission rate, mortality rate, age and sex ratio, symptoms, incubation, how to protect yourself, myth busters, virus life on different surfaces


Rajanpandey:
Refer to the following websites.
1. https://www.cov19.xyz/info/mortality-rate
2. https://worldcoronavirus.org/symptoms

We can use the same information and put it on this website.

sudevschiz:
All this could probably go into a "Deep-Dive" tab. Could be a nice separate project on its own.

PR welcome.

shuklaayush:
Relevant: #141

sagarv1997:
Just read the contributing guidelines @sudevschiz Please assign the Myth Buster section to me

jeremyphilemon:
Added by #539, #325 addresses myth busters

PR #228, opened 2020-03-27
Population :: Add a new column against each state mentioning the popu…

### LINKED ISSUE #209: Population :: Add a new column against each state mentioning the population or Cases/1M population
Population :: Add a new column against each state mentioning the population or Cases/1M population.

This will help understand how severe hit a state is.

iam-hitesh:
Fixes #209 

Added population.json file for getting population of different states and union territories, census data is taken from 
https://data.gov.in/resources/stateut-wise-overall-population-st-population-percentage-sts-indiastate-total-population


iam-hitesh:
Added new column to table ie. cases/1M.
Please review and merge. And let me know if there any changes.

abhinav098:
UI breaks on my mac 13" screen. 

Try making some change to the CSS.
.Home {
    display: flex;
    flex-direction: row;
    justify-content: space-around;
    flex-wrap: wrap;
    margin-left: 5rem; # earlier it was 10rem
    margin-right: 5rem; # earlier it was 10rem
}

refer screenshot for the issue => 

Screenshot 2020-03-27 at 23 16 03

shuklaayush:
Hi @iam-hitesh, see #221. I don't think we have enough space on the table to show this as well. The UI breaks on smartphones. It is an interesting metric, though. Maybe we can add it in a deep dive section (#141, #190). Closing.

JunaidBabu:
We can add this later in our deep dive tab (in-progress)

PR #237, opened 2020-03-27
Revamped Charts 

### LINKED ISSUE #10: Add Exponential Toggle on the Charts
Append a mechanism to **view the difference** (and thereby, exponentiation in growth) in the number of cases in comparison to the day preceding the one over which the user hovers.

This presents a 'trend spread,' which makes the content more informative and useful.

### LINKED ISSUE #136: Log graphs
Covid-19 is an exponential disease.

Looking at a simple graph really does not give true picture of cases.

Log graph will give correct picture on whether the disease is slowing down or further accelerating.

If possible, have log graph on a per state basis

### LINKED ISSUE #189: Change the Daily trends to bar graph

s-naveen:
Fixes #136 
Fixes #10 
Fixes #189


src/components/charts.js
```diff
@@ -0,0 +1,237 @@
...
+  defaults.global.legend.display = true;
+  defaults.global.legend.position = 'bottom';
+
+  defaults.global.hover.intersect = false;
+
+  if (!props.timeseries || props.timeseries.length == 0) {
+    return <div></div>;
+  }
+
+  props.timeseries.forEach((data, index) => {
```
vramana, line 38: Why not destructure timeseries from props? Then you stop writing props everywhere. It does not add any value.

```js
const { timeseries } = props
```
or even better
 
```js
function Charts({ graphOption, mode, timeseries })
```

jeremyphilemon:
Hey so it was established that we'd be sticking to the current d3 charts, however, it does look like we need a logarithmic switch. I'll be opening an issue to discuss over the graph enhancements, we'll have a discussion and move to the PRs :)

PS: Those 12 commits train tho :P

PR #251, opened 2020-03-27
Added GeoJSON for missing Andhra and Telangana

PrasannaBrabourame:
Andhra Pradesh and Telangana GeoJson Added.
one concole.log is removed


jeremyphilemon:
Thanks!

PR #256, opened 2020-03-28
Fix district maps by adding district aliases (fixes #253)

### LINKED ISSUE #253: Issues with district names
**Describe the bug**
District names of some districts are different in the maps and the data that we have. E.g Bangalore, Bengaluru; Mysore, Mysuru

**To Reproduce**
1. Go to site
2. Click on Karnataka in the map
3. See that neither of Bangalore or Mysore are coloured and have 0 cases

**Expected behavior**
They have more than zero cases and hence must be coloured

**Screenshots**
Map of Karnataka
image

**Additional context**
I don't know what is the preferred approach for fixing this issue, fixing the map files or creating an alias map. The alias map might be useful in case we don't have data for the exact district but for a group. (I am not sure about this, but is the data for Bangalore and Bangalore Rural combined or separately available?)

Nevertheless, I will try to create a PR for districts that have been renamed or reorganized if I can find them or find updated maps

meghprkh:
Temporary measure till we have more up-to-date maps (was unable to find them)

Though most of these are just alternate spellings

Fixes #253

### DIFF
--- a/src/components/district_aliases.js
+++ b/src/components/district_aliases.js
@@ -1,0 +1,52 @@
+// Temporary measure till we have more up-to-date maps
+// Though most of these are just alternate spellings
+
+const DISTRICT_ALIASES = {
+    "Delhi": {
+        "East Delhi": "East",
+        "South West Delhi": "South West",
+        "West Delhi": "West",
+        "Delhi": "New Delhi",
+        "South Delhi": "South",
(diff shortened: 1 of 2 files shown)


meghprkh:
Sorry for all the force pushes. Everything seems to work now :)

JunaidBabu:
Fixed issue at geojson

PR #264, opened 2020-03-28
lavanya | Add Total confirmed count for the state with districts view

lavac:


shuklaayush:
Hi @lavac. Thanks for the PR, but I think it's redundant to have both the state count, as well as a separate total count in the district view. Closing for now. Will reopen if the need arises.

PR #265, opened 2020-03-28
Remove default uniform scaling

dhruvdutt:
The graphs are getting obscure because of wrong scale. Please let's not scale uniformly. This is much better picture for death and recoveries:

image

Same for **Daily**. At first look, it seems like nothing is changing much but in real the spikes are huge (check last screenshot below):

image

image

### DIFF
--- a/src/components/home.js
+++ b/src/components/home.js
@@ -18,7 +18,7 @@ function Home(props) {
   const [lastUpdated, setLastUpdated] = useState('');
   const [timeseries, setTimeseries] = useState([]);
   const [deltas, setDeltas] = useState([]);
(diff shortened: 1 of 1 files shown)


dhwanilvyas:
Looks cleaner this way, the difference is huge but setting scaling uniformly by default makes it a bit unclear at first.

JunaidBabu:
Thank you for the PR. But uniform scaling is set to default by a design choice.

Closing this PR

dhruvdutt:
@jeremyphilemon @JunaidBabu 
I assumed that it must a design choice but can we please have some discussion before closing the PR about why such design choice was made?

Issue #252, opened 2020-03-28
Use redux or react context for state management.


rohitpotato:
After seeing the source code, it is quite obvious that a lot of prop drilling is going on and functions are being passed from component to component. This will become a lot trickier when the project grows. 

I sugggest a battle tested state management solution for the project. Redux or Context maybe?

JunaidBabu:
@jeremyphilemon @shuklaayush @s-naveen thoughts?

shuklaayush:
I don't have any experience with either but would love to learn.

rohitpotato:
Can i make a pull request with a redux setup for the project?

s-naveen:
I am not that much familiar with redux either. Maybe we can try it if it does not make codebase complex.

rohitpotato:
> I am not that much familiar with redux either. Maybe we can try it if it does not make codebase complex.

The codebase will get trickier if the project expands and there's no state management container in place.

findelallc:
Guys I don't think it will require a redux wrapped for this project as of use case, a plan to use react context in one of the main part would be sounds good to prevent memory leaking..Don't make the codebase complex for the sake of simplicity and usability.

rohitpotato:
Hi everyone, since I am new to open source contribution on github, I didn't exactly see the guidelines and made a pull request. Really sorry for that. Would anyone mind reviewing that?

naveen-nrj:
I think redux using hooks will not be that much complex since we are using functional components
So redux with hooks is a good choice. Also, Don't know much about context.

vramana:
Hi, I just came across this repo. I have experience with Redux and Context. I'll take a look around source and try to make suggestion.  But simply sticking to prop drilling is just fine in some cases even if it's 5 to 6 levels deep.

rohitpotato:
@naveen-nrj @KrishnaPravin  check out a pull request I made: [HERE](https://github.com/covid19india/covid19india-react/pull/284)

rohitpotato:
> Hi, I just came across this repo. I have experience with Redux and Context. I'll take a look around source and try to make suggestion. But simply sticking to prop drilling is just fine in some cases even if it's 5 to 6 levels deep.

Prop drilling is just bad practice and it becomes harder to debug when the project grows. I am assuming this project will eventually grow. It makes a lot of sense to stick with battle tested solutions.

vramana:
[...]

PR #284, opened 2020-03-28
Added Redux Setup for the project

rohitpotato:
After seeing the source code, it is quite obvious that a lot of prop drilling is going on and functions are being passed from component to component. This will become a lot trickier when the project grows.

I sugggest a battle tested state management solution for the project. Redux or Context maybe?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15565,6 +15565,18 @@
         "prop-types": "*"
       }
     },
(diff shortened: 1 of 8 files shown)


rohitpotato:
For #252

jeremyphilemon:
Please check comments in issue #252

Issue #247, opened 2020-03-27
Active Recovered Deceased count is not displayed after state is selected


gosharmaa:
**Describe the bug**
Active Recovered Deceased count is not displayed after state is selected

**To Reproduce**
1. Go to Maharashtra in the map
2. Click on different cities 
3. Notice that Active Recovered Deceased count is not getting updated

**Expected behavior**
Information should get updated correctly

**Screenshots**
![Screenshot_20200328-015015](https://user-images.githubusercontent.com/62764438/77797236-bdcdae00-7096-11ea-92c3-17b98fad99b8.jpg)

jeremyphilemon:
The district map support was a recent feature addition! We're only showing "confirmed" data as of now hence the "NA", but once we're sure with our datapoints at a district level granularity, we'll add those datapoints.

@JunaidBabu @sudevschiz word on when we'll be supporting district datapoints?

JunaidBabu:
> @JunaidBabu @sudevschiz word on when we'll be supporting district datapoints?

Not any time soon. We should remove Active/Recovered/Decease after selecting state

anilgupta1591:
How can contribute or how can take ticket ?
Sorry i know this is not proper medium to ask this question but dont know any other source for this like slack channel

puneetnsit:
Working on it

gaurav7118:
@jeremyphilemon
I have solved the issue locally but not able to create review or push to master. It is showing unable to access master.

JunaidBabu:
This cannot be fixed because we do not have the data at district level. Closing this issue.

jeremyphilemon:
@anilgupta1591 head over to the issues section, pick any accepted issue and open a PR :)

PR #286, opened 2020-03-28
Fix: Show only confirmed values for district level

puneetnsit:
https://github.com/covid19india/covid19india-react/issues/247

Applied checks on mapexplorer.js. Now Active,Recovered,Deceased wont appear for state level


jeremyphilemon:
Hey thanks for the PR! The issue seems to be closed now, and even so I believe there should be a different strategy at hand design wise, since the box tends to fill the whole row once we're through district level.

This will be discussed in a future issue and hopefully we can have a discussion on potential strategies before we have people open PRs. Thank you!

PR #288, opened 2020-03-28
Highlight district on Map while hovering table

### LINKED ISSUE #281: Districts not highlighted on map while hovering table
As in the case of states, districts should also be highlighted while hovering over their table entries.

shivanshs9:
Fixes #281

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -521,7 +521,7 @@ table {
         background: $gray-hover;
       }
       &:nth-child(odd) {
-        background: $white !important;
+        background: $white;
       }
     }
 
@@ -588,6 +588,14 @@ table {
       transform: translateX(1rem);
       .unknown {
         font-style: italic;
+      }
+
+      &:nth-child(odd) {
+        background: $light !important;
+      }
+
+      &:hover {
+        background: $gray-hover !important;
       }
     }
   }
--- a/src/components/home.js
+++ b/src/components/home.js
@@ -60,16 +60,7 @@ function Home(props) {
   const onHighlightDistrict = (districtName, index) => {
     if (!districtName && !index) setMapRegionHighlighted(null);
     else {
-      const district = {
-        districtData: {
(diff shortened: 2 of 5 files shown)


shuklaayush:
Hey @shivanshs9, thanks for this! ~~Seems to work great~~. But please revert the extra formatting changes before we can merge. We're keeping the formatting changes for a separate PR.

Edit: Found two small bugs on further analysis:
- Expanding a state map still allows other states to show up on the header at the top of the map. This could be misleading.
Peek 2020-03-28 22-36

- Pratapgarh (Rajasthan) seems to highlight under UP since there's another Pratapgarh district in UP as well.
Peek 2020-03-28 22-32

We'll need to add a check somewhere to ensure that **only** districts under the current state are highlighted and displayed on the header when a state map is expanded. This might not be trivial to fix. Let me know if you need any help.

jeremyphilemon:
@shuklaayush also thoughts on opening the state map once the district tables are open, and then hover over districts starts highlighting the districts. @shivanshs9 Let us know if you can do this after the bug fixes!

Otherwise this looks good!

shuklaayush:
@jeremyphilemon UX-wise, that's what makes the most sense. We just have to think about what happens if multiple states are expanded. Should we just expand the map to the last state? What if the last one is closed, do we revert to the earlier one or to the whole India map? Or should we just not allow multiple states expansion in the table? Let's iron out the details first and then start implementing.

shivanshs9:
@shuklaayush If it's needed, I can undo the formatting changes. Right now, I've cleared the formatting changes conflicting with the master changes. But I haven't touched other changes yet. Let me know if I need to push a commit to undo that.

PS: I've fixed both the bugs you mentioned. :+1:

jeremyphilemon:
Hey thanks for working on this! A later PR had made a patch fixing this! Thank you :)

PR #295, opened 2020-03-28
Replaced ES6 Map with lodash map for performance

keshavbiswa:
Refactor:

Lodash will enhance performance as the app scales.
Replaced all ES6 maps to lodash maps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "date-fns": "^2.11.0",
     "date-fns-tz": "^1.0.10",
     "install": "^0.13.0",
+    "lodash": "^4.17.15",
     "nice-react-ticker": "^1.1.0",
     "node-sass": "^4.13.1",
     "npm": "^6.14.3",
--- a/src/components/level.js
+++ b/src/components/level.js
@@ -1,4 +1,5 @@
 import React, {useState, useEffect} from 'react';
+import _ from 'lodash';
 
 function Level(props) {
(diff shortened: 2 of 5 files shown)


crup:
`import _ from 'lodash';`

I don't think this will treeshake properly. If I am right, this PR is bloating the bundle by 30kb+

Rather try `import map from from "loadsh/map"`

https://www.covid19india.org/
image

https://covid19india-react-git-fork-keshavbiswa-integrate-lodash.covid19india.now.sh/
image

vramana:
> Lodash will enhance performance as the app scales.
> Replaced all ES6 maps to lodash maps.

A lot of performance has gone into Array.prototype.map performance over the last few years. How much difference is does it make? Can you run some benchmarks and see if there is percievable difference? If it's 10ms vs 15ms It's not that much. But 1sec vs 3sec that's a much bigger difference. The general rule to make any performance improvements is measure the thing you want to improve.

jeremyphilemon:
Hey, this is an interesting PR! Honestly I don't see much impact on switching to lodash for very minimal performance increases, given the arrays are not that large. 

If you could run some benchmark tests and share the results, you can prolly change my mind :)

keshavbiswa:
Now that I think about it. Doesn't really make any difference and the data is always going to remain small so doesn't have any impact. Going to close this pr now.

Issue #145, opened 2020-03-26
Toggle switch for 'Scale Uniformly'


shuklaayush:
The checkbox is difficult to tick on mobile. Adding a toggle switch instead would help. 
PS: Maybe change the text to 'Uniform Scaling' as well.

VigneshMurugan:
Will pick this up

sivasatvik:
Made a PR  #158 for this. It works for desktop and mobile versions. Suggestions?

jeremyphilemon:
DUNZOOOO THANKS TO @sivasatvik!

PR #311, opened 2020-03-28
Add a toggle instead of a checkbox

sivasatvik:
A checkbox might be inconvenient for mobile users, hence adding a toggle instead. Solves the issue #145 

Annotation 2020-03-26 133008


jeremyphilemon:
Awesome I loooove the toggle haha! Almost surprised that all it took was some css! Thank you!

PR #328, opened 2020-03-29
Fix react-hook related warnings

### LINKED ISSUE #313: Fixing react-hooks/exhaustive-deps warnings.
**Describe the bug**
Would be great if we could fix the `react-hooks/exhaustive-deps` warnings shown in the console . 

**To Reproduce**
Just open your console after `npm I && npm start`

**Expected behavior**
The console should be clear of these warnings.

**Screenshots**
Screen Shot 2020-03-29 at 8 14 54 AM

**Additional context**
Fixing this would mean reading up `useEffect` and how their dependencies work. Here's a nice little guide https://overreacted.io/a-complete-guide-to-useeffect/

ssonal:
Closes #313

### DIFF
--- a/src/components/banner.js
+++ b/src/components/banner.js
@@ -1,60 +1,36 @@
-import React, {useState, useEffect} from 'react';
-import differenceInMilliseconds from 'date-fns/differenceInMilliseconds';
+import React, {useState, useEffect, useCallback} from 'react';
 import axios from 'axios';
 
 function Banner(props) {
   const [snippets, setSnippets] = useState([]);
   const [snippet, setSnippet] = useState();
-  const [start] = useState(
-    props.start
-      ? new Date(props.date)
-      : localStorage.getItem('start') === 'null'
-      ? new Date()
-      : new Date(localStorage.getItem('start'))
-  );
-  const [difference, setDifference] = useState(
-    new Date(differenceInMilliseconds(new Date(), start))
-      .toISOString()
-      .slice(11, 19)
-  );
 
   useEffect(() => {
-    getSnippets();
-  }, [1]);
-
-  useEffect(() => {
-    if (snippets.length > 1) {
-      setSnippet(snippets[0]);
-    }
-  }, [snippets]);
-
-  useEffect(() => {
-    const interval = setInterval(() => {
-      setDifference(
-        new Date(differenceInMilliseconds(new Date(), start))
-          .toISOString()
-          .slice(11, 19)
-      );
-    }, 10000);
-    snippetChooser(0, snippets.length - 1);
-    return () => clearInterval(interval);
-  }, [difference]);
-
-  const snippetChooser = (min, max) => {
-    const index = Math.random() * (max - min) + min;
-    setSnippet(snippets[Math.floor(index)]);
-  };
-
-  const getSnippets = () => {
     axios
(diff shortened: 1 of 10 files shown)


ssonal:
All warnings now cleaned up.

.prettierignore
```diff
@@ -0,0 +1 @@
+public/maps/
```
vinay-thallam: @ssonal 

It is better to prettify the maps JSON also. There can be partial modifications to maps json in future, in that case diff does not show properly if not prettified. 

Below is the example of such commit done by one of the contributors,

https://github.com/covid19india/covid19india-react/pull/283/commits/1bb93ee02f52943ad5b88ffa2e8333a0b8be5a32

vinay-thallam: Better way to speed up the pre-commit checks is to run the scripts only on the staged files instead of all files in the project using "lint-staged" configurations.

src/components/banner.js
```diff
@@ -38,12 +46,7 @@ function Banner(props) {
     }, 10000);
     snippetChooser(0, snippets.length - 1);
     return () => clearInterval(interval);
-  }, [difference]);
-
-  const snippetChooser = (min, max) => {
-    const index = Math.random() * (max - min) + min;
-    setSnippet(snippets[Math.floor(index)]);
-  };
+  }, [difference, snippetChooser, snippets.length, start]);
```
vinay-thallam: Is the "difference" really needed in the dep array for this effect ? Its value is not consumed in the effect.  
Also setDifference is being called which changes difference and results in effect being triggered again and so on so forth.

@jeremyphilemon  @ssonal

src/components/banner.js
```diff
@@ -20,14 +20,22 @@ function Banner(props) {
 
   useEffect(() => {
     getSnippets();
-  }, [1]);
+  }, []);
 
   useEffect(() => {
```
vinay-thallam: This effect is not needed. 
Instead below block of code can be put at line number 53 i.e inside axios success callback. 

`if (response.data.factoids.length > 1) {
      setSnippet(response.data.factoids[0]);
    }`

src/components/banner.js
ssonal: Makes sense. Fixed, thanks!

src/components/banner.js
ssonal: Thanks, fixed!

jeremyphilemon:
We added a little bit more tweaks to the choropleth map in a previous patch, once you're done with resolving the conflicts we're good to go :)

.prettierignore
jeremyphilemon: Also it looks like I'm having trouble getting lint-staged to work? I've added the configuration but somehow when staging the files don't seem to lint, I'm not sure if I'm doing it wrong or if there's an issue. If you'd like to look into it, feel free to do so!

vinay-thallam: @jeremyphilemon  Addressed it in the below PR
https://github.com/covid19india/covid19india-react/pull/364

Please review

ssonal:
@jeremyphilemon rebased this branch on to master.

.prettierignore
ssonal: Dropped this commit from my PR. Best to handle it separately.

PR #334, opened 2020-03-29
Lazy loading and custom hook for managing data in Home screen

vamshi9666:


vramana:
What's the percievable difference here? Have you measured any numbers? Lighthouse score? Time to load the page, initial bundle size etc.,

vamshi9666:
without lazy loading : 
Screenshot 2020-03-29 at 8 00 37 PM
with lazy loading : 
Screenshot 2020-03-29 at 7 59 59 PM

vramana:
@vamshi9666 Seems like it knocks off 40kb. Can you measure more things? What happens on fast & slow 3G connections? 

Measure impact of on the time. Bundle size is only part the equation probably the easiest thing to measure.

src/App.js
```diff
@@ -1,60 +1,77 @@
...
 import './App.scss';
-import Home from './components/home';
-import Navbar from './components/navbar';
-import Links from './components/links';
-import Summary from './components/summary';
-import Cluster from './components/cluster';
-import FAQ from './components/faq';
-import Banner from './components/banner';
+const Home = lazy(() => import('./components/home'));
+const Navbar = lazy(() => import('./components/navbar'));
```
vramana: Navbar is needed for all pages. So why should it be split into chunk?

vamshi9666: it may not need to be lazy.I just multi selected and wrote.

vamshi9666: made it non-lazy

vamshi9666:
any more suggestions ?

vamshi9666:
added commits for seperate hook for managing data in Home

vamshi9666:
@jeremyphilemon @JunaidBabu   Please review

jeremyphilemon:
Hey thanks for the PR! As mentioned in the [contributing](https://github.com/covid19india/covid19india-react/blob/master/CONTRIBUTING.md) guide, please have an issue filed first, have it accepted, then assigned to you before submitting a PR :)

This will help us discuss any changes/requirements before work is done on it!

Issue #355, opened 2020-03-29
Test Coverage


tannakartikey:
**Describe the solution you'd like**

The project is attracting lots of contributors and the data we display is critical. So I thought if we add some test coverage to the codebase it will be great. That way we will be sure that the calculations are correct, new changes are not breaking anything and we will be confident while refactoring... (you know the benefits of the tests)

Since our project is open source we can use any CI/CD tool like Travis or use Zeit's build step.

shuklaayush:
Yes! We need something other than just prettier errors to fix when pushing.

jeremyphilemon:
Closing this because we now support tests. Henceforth we shall open new issues for adding specific test cases. 

@tannakartikey you can recommend, we can assign it to people looking to contribute :)

PR #356, opened 2020-03-29
#355 Initial Test Setup

tannakartikey:
Following things are included in this PR:
 - Tests setup and configuration
- A spec to calculate Level component.
 - A spec that downloads state data from API and then compares the data with the map files (`public/maps/*`) and reports discrepancy if any. It fails on the first error it finds instead of reporting errors in all the states. This limitation is due to https://github.com/facebook/jest/issues/1619
 - Another spec runs dynamically for each state against the static data downloaded from the API and extracted from the maps file. This test is helpful right now to see where the data has discrepancies. Once fixed we can get rid of this test
 - Under `testUtils/index.js` there are two functions that return an object with state names as keys and cities as values using the API and map files(`public/maps/`).
 - Some data cleaning. A few of the map files did not follow the convention the same as other files. Now they all follow the same standards for storing the cities data. For example, if the file name is `gujarat.json` then the cities will be found under `gujarat_district` property so on and so forth.

Command to run the specs:
`yarn test`

Ref: #355


jeremyphilemon:
If you're working on adding tests, that would be super cool! I'll be keeping an eye on this PR :)

jeremyphilemon:
Just a clarification, is this a WIP or is it ready for a merge?

tannakartikey:
It has the setup for tests and one sample test. We can merge this and
include more specs in subsequent PRs.

tannakartikey:
@jeremyphilemon I have changed some configuration of the tests.

I saw that we constantly getting issues on the difference of the name in API response and the map data.

So I have added a test that downloads the data from this API endpoint, https://api.covid19india.org/state_district_wise.json, and compares it with the data stored inside `public/maps`.

One drawback is that it will stop when the assertions don't match. So, instead of showing all the errors of all the states it will stop on the first state where it fails. You can see the result in the screenshot. The test will move further only when we fix that state.

I can do the workaround by storing the static data. But then if it ever changes we need to update the spec manually.

image

tannakartikey:
I downloaded the data and implemented dynamic specs based on that static data. So, at least for now, we know what states need improvements in the data. Later when all the specs are green we can get rid of the static specs.

image

Command to run this spec: 
```
yarn run test --testPathPattern stateNames.test.js
```

tannakartikey:
@jeremyphilemon I think we can merge this PR here and then keep creating new PRs as we go on writing specs.

tannakartikey:
@jeremyphilemon GitHub action added to run the specs on each push and PR.

JunaidBabu:
Is it possible to simplify this PR by having a very basic test environment? It's very difficult to review this PR with so many changes.

And I think the data discrepancy tests should be done on the https://github.com/covid19india/api repository.

Closing this PR now. Please raise an issue regarding Tests (if it doesn't exist already) and discuss how to create tests before submitting.

tannakartikey:
@JunaidBabu I have created a new PR with an only basic setup: https://github.com/covid19india/covid19india-react/pull/960

I had included the data discrepancy test in this PR because we were getting lots of issues related to data. Issue #355 can be the place to discuss further about data discrepancy tests. I will add comments there too.

Meanwhile, again, here is the new PR: https://github.com/covid19india/covid19india-react/pull/960

Issue #346, opened 2020-03-29
Ahmedabad district misspelt in raw data; hence cases not showing on the map


mmshah278:
**Describe the bug**
0 cases of Ahmedabad city/district's are shown on the map when there are 20 (when this is being written)

**To Reproduce**
Go to and click on the Gujarat portion of the map and then go to and hover on the Ahmedabad city portion of the map. 0.00% cases from Ahmedabad will be displayed.

**Expected behavior**
Instead 0.00%, 31.74% should be displayed when hovered on Ahmedabad district

**Desktop (please complete the following information):**

- OS: win10
- Firefox
- 74.0

**Additional context**
The raw data contains "Ahmadabad" as district name for Ahmedabad city under the "Detected District". Cleaning it might remove this error. Second place to look for this correction is in "district_list"

rodiwa:
i checked on https://api.covid19india.org/raw_data.json and looks like it needs to be 'ahmadabad' (with a 'a'). Here's a quick fix PR, if that's the case - https://github.com/covid19india/covid19india-react/pull/358.

mmshah278:
@rodiwa Actually, shouldn't it be "Ahmedabad" with an 'e' because that is what the name of the city and district is?

rodiwa:
Totally agree :) But I see that the data has already been setup that way. This quick-fix basically starts showing up the count on the map.
PS - To fix to "Ahmedabad", changes to https://api.covid19india.org/raw_data.json will be needed and we'll need to ensure new data are now vary of this update.

karan1919:
similar issue exists for Ahmadnagar, Maharashtra; cases show correctly but district is not highlighted.

RajendrasinhParmar:
This was previously working just yesterday it was showing correct values and highlights on map for **Ahmedabad**.

Is there anything that is updated in source data somehow? and how much control do we have over source data?

shuklaayush:
Fixed via #429

PR #358, opened 2020-03-29
COVID-346-fixed-spelling

rodiwa:
quick fix for https://github.com/covid19india/covid19india-react/issues/346


bizzappdev:
Ahmadabad is no a correct name. instead Ahmedabad has to be added into the API. 
https://github.com/covid19india/api/pull/54

shuklaayush:
Thanks, but this seems to have been fixed via #429

PR #361, opened 2020-03-29
Make map scaling/centering automatic

shuklaayush:
Removed hardcoded values of center/scale for each map json. This ensures that all maps when expanded are about the same size and would make it easier if map files are ever changed. Also, less code to maintain.

Inadvertently, also fixes map cropping in landscape orientation on mobile.


jeremyphilemon:
> Inadvertently, also fixes map cropping in landscape orientation on mobile.

Ahh when one fix kills two bugs

PR #364, opened 2020-03-30
Fix lint staged

vinay-thallam:
Modified lint-staged config to run pre-commit hook scripts on staged files only


shuklaayush:
Fixes #369

PR #366, opened 2020-03-30
[COVID-346]:Fix spelling mistake in ahmednagar maharashtra

pk1997:
Change spelling of ahmednagar in MH


shuklaayush:
Thanks, but this seems to have been fixed via #429

Issue #369, opened 2020-03-30
Fix all lint errors


pk1997:
Fix all lint errors since linter isn't allowing to commit because of them

shuklaayush:
@jeremyphilemon Please.

pk1997:
I have put a PR for this.Please review and merge

pk1997:
Fixed in PR #370

PR #370, opened 2020-03-30
[COVID-369]: Fix all lint errors

pk1997:


shuklaayush:
Fixes #369. Though I think this is already covered in #364.

shuklaayush:
@pk1997 Hey! Since this was fixed by an earlier PR, we merged that one. Thanks!

PR #373, opened 2020-03-30
Fix the Ahmadabad spell

ajaypatelaj:
Due to mismatch of spelling Ahmadabad in data and Ahmedabad in map.  Gujarat map is not showing Ahmadabad data.
I have set the same spelling as data for map, it should fix the issue. https://github.com/covid19india/covid19india-react/issues/346


shuklaayush:
Fixed via #521

PR #383, opened 2020-03-30
[Ashish] fix district name for Pauri Garhwal in Uttarakhand

IMMarkXLII:
name in maps/uttarakhand.json is incorrectly spelled Garhwal for district Pauri Garhwal


guneetnarula:
Related to #413

shuklaayush:
Fixed via #521

PR #389, opened 2020-03-30
fixes the issue on click of total in table

### LINKED ISSUE #387: Throws an error after clicking on total in the state table
**Description **
Throws an error after clicking on Total in the table state table
you can see the error in the JS console. Refer to the screenshot.

**To Reproduce**
1. Go to the [https://www.covid19india.org/](https://www.covid19india.org/) 
2. open JS console.
2. Click on Total in State table 

**Expected behavior**
Should not throw any error.

**Screenshots**
Screenshot 2020-03-30 at 4 00 30 PM

rakeshreddy1410:
Fixed #387 

Shouldn't throw any errors in console


shuklaayush:
Nice catch!

src/components/row.js
```diff
@@ -94,7 +94,9 @@ function Row(props) {
         onMouseLeave={() => props.onHighlightState?.()}
         touchstart={() => props.onHighlightState?.(state, props.index)}
         onClick={() => {
-          handleReveal();
+          if (!props.total) {
+            handleReveal();
+          }
         }}
```
chinmaydalvi: @rakeshreddy1410  
Can we avoid the attaching of a click event on `Total` as there is no use of it?
Due to your changes click event still will fire after clicking on it only thing  is you are avoiding the call of `handleReveal()` function.

I fill that this is not the best way. because there is no use of click event on `Total Bar` so instead of preventing it by adding the condition better don't attach event on `Total bar` 

@shuklaayush  Your opinion also matters.

shuklaayush: @chinmaydalvi You're right. No point in even attaching a click event to the Total row

shastri48: @chinmaydalvi  An alternate way would be to pass and spoof function to Total in the `Table.js` file.
Like: 
```suggestion
       {states.length > 1 && props.summary === false && (
            <Row key={0} state={states[0]} total={true}  handleReveal={() => false}/>
          )}
```

rakeshreddy1410: since we are doing conditional validation in row.js which already gets props.total
added the event based on the same condition 
updated the pull req

src/components/row.js
```diff
@@ -93,9 +93,7 @@ function Row(props) {
         onMouseEnter={() => props.onHighlightState?.(state, props.index)}
         onMouseLeave={() => props.onHighlightState?.()}
         touchstart={() => props.onHighlightState?.(state, props.index)}
-        onClick={() => {
-          handleReveal();
-        }}
+        onClick={!props.total ? handleReveal : null}
```
rakeshreddy1410, line 100: conditional event based on the input props

chinmaydalvi: In your solution, you are still attaching an event. I was saying don't attach event if it is total row`

Refer my solution:

You can do following on line no 77 
`const clickEvent = !props.total ? { 'onClick': handleReveal } : {};`

and then on line no 96 you can write  ` {... clickEvent}`
```
<tr className={props.total ? 'state is-total' : 'state'}
      onMouseEnter={() => props.onHighlightState?.(state, props.index)}
      onMouseLeave={() => props.onHighlightState?.()}
      touchstart={() => props.onHighlightState?.(state, props.index)}
      {... clickEvent}
>

```

chinmaydalvi: Also, Please resolve conflicts

jeremyphilemon:
@shuklaayush is this solved?

@rakeshreddy1410 if so, you could fix the conflicts and rebase? :)

shuklaayush:
@jeremyphilemon Not on the current master. It still throws an error locally.
image

PR #394, opened 2020-03-30
   feat: Included covid-19 self assessment app

sanjaybxl:
# How you feel - covid-19 self assessment app

Because of the fact that the Covid-19 is more and more present, and it’s very similar to other infections such as flu or cold, we’d like to provide people with a web-based self-assessment tool. How You Feel app allows you to perform self-checkup to determine the possibility of SARS-Cov-19 infection, as well as a cross-reference for alternative reasons for the combination of symptoms. It gives the user a number of recommendations one should take into consideration in the nearest future.

Couple things:

* The app will not track - ever - do not ask for adding that feature.
* The app will not send the symptom data anywhere

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,6 +12,7 @@ import Home from './components/home';
 import Navbar from './components/navbar';
 import Links from './components/links';
 import Cluster from './components/cluster';
+import HowYouFeel from './components/howyoufeel';
 import FAQ from './components/faq';
 import Banner from './components/banner';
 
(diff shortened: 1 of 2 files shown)


jeremyphilemon:
:/ don't see anything. 

As mentioned in the [contributing](https://github.com/covid19india/covid19india-react/blob/master/CONTRIBUTING.md) guide, please have an issue filed first, have it accepted, then assigned to you before submitting a PR :)

This will help us discuss any changes/requirements before work is done on it!

PR #400, opened 2020-03-30
Fixed react-hooks dependency lists errors and warnings

jeremyphilemon:
- Thanks to @ssonal!

Also there seemed to be a problem with the minigraphs rendering. Having `timeseries`, `Graphdata`, and `time series.length` in the dependency list redrew the d3 graphs multiple times. I've disabled that for now so if you can come up with a fix for it next, we can start enforcing the react-hooks rules in the eslint.js files :)

### DIFF
--- a/src/components/minigraph.js
+++ b/src/components/minigraph.js
@@ -1,4 +1,4 @@
-import React, {useState, useEffect, useRef} from 'react';
+import React, {useState, useEffect, useRef, useCallback} from 'react';
 import * as d3 from 'd3';
 
 function Minigraph(props) {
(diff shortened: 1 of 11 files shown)


src/components/minigraph.js
```diff
@@ -29,274 +23,283 @@ function Minigraph(props) {
...
-      })
-      .style('opacity', 0)
-      .transition()
-      .duration(props.animate ? 3000 : 0)
-      .style('opacity', 1);
-  }
+  useEffect(() => {
+    graphData(timeseries);
+    // eslint-disable-next-line
+  }, [timeseries]);
```
jeremyphilemon, line 302: @ssonal here.

ssonal: Just patched this in #402. Should work now!

Side-note, testing these conditions is pretty painful. How are you doing it?

jeremyphilemon: Haha I feel you, it's a pain! 

So I noticed the graphs looked odd, so inspected the code and added a `console.log` in the useEffect to see how many times the state changes happened in the dependency list, and voila! The graphData was called so many times having it redrawn so many times.

Could've slipped my eyes, but hey :P

jeremyphilemon: @ssonal also I had a look at the time series graphs, I think it's the same issue. The graphs tends to be redrawn again. 

See if you can fix it too!

PR #401, opened 2020-03-30
Replace sorting symbols with triangles

shuklaayush:
Less ambiguous.

image


jeremyphilemon:
You wanna add it to district table too?

shuklaayush:
Oops. Added :+1:

PR #407, opened 2020-03-30
Addition of Helpline Numbers for every state

bhaskarghosh16:
A quick view of helpline numbers for every state in the India map could be extremely helpful in times of need. 
Information gathered from : https://www.mohfw.gov.in/coronvavirushelplinenumber.pdf


jeremyphilemon:
Commented in #425

Issue #409, opened 2020-03-30
Uttarakhand Map does not show information for Pauri Garhwal district


tuxdna:
Steps to reproduce are same as in https://github.com/covid19india/covid19india-react/issues/301

However there are two issues:

 * The map file for Uttarakhand state was reverted ( Json pretty formatted ), and district Pauri got reverted back to Garhwal 
 * The correct name should now be "Pauri Garhwal" as in the Google Spreadsheet (rows 640, 921 ) - https://docs.google.com/spreadsheets/d/e/2PACX-1vSc_2y5N0I67wDU38DjDh35IZSIS30rQf7_NYZhtYYGU1jJYT6_kDx4YpF-qw0LSlGsBYP8pqM_a1Pd/pubhtml

 
We need to fix the map to reflect correct district name to "Pauri Garhwal" now.

shuklaayush:
Duplicate of #404

PR #410, opened 2020-03-30
Fix: Uttarakhand Map does not show information for Pauri Garhwal 

tuxdna:
Fixes issue #409


shuklaayush:
Thanks for this, but I think it's already fixed in #383

Issue #416, opened 2020-03-31
Add the UPI / Account information of PM and other relief funds.


dpnkrpl:
It would be great if we urge people to contribute some amount through this website .
state wise relief funds information should be added in the helpful links and `pmcares@sbi` UPI should be added in the banner.

For mobile users we could also redirect them to a UPI portal ( if possible, i'm not sure ).

Attention: according to Wikipedia ( https://en.wikipedia.org/wiki/PM_CARES_Fund) 
"While the original UPI account was PMCARES@SBI, Delhi Police booked unknown individuals for creating an account removing the 'S', PMCARE@SBI, intended to scam people." so we should warn people too !!

atharvaj77:
Will work on it

pankaj-dahiya-devops:
I suggest you to add CM relief fund details too, so that people who want to donate to CM relief fund can do so.

jeremyphilemon:
Closed by #619

PR #340, opened 2020-03-29
Add Donation Link to PMNRF in Helpful Links

rajhrikshit:

PR #419, opened 2020-03-31
Added info about PM cares fund

atharvaj77:
With regard to #416 added info of PM cares fund and also added few images in public directory


shuklaayush:
Similar to #340

Issue #413, opened 2020-03-30
Update all district shapefiles (JSONs)


guneetnarula:
We are using old district boundaries for most states. Telangana, West Bengal, Manipur, Gujarat, Maharashtra to name a few. This is causing both name mismatch problems and missing data

**Solution**
We have access to district boundaries from 2019, we can replace the JSONs we are using with new ones. The new JSONs also match well with the district names reference we are using for data collection.

Should resolve related issues like #253 and #256

harshvitra:
Can you share the json files we can fix this easily

guneetnarula:
Yes, on it

guneetnarula:
Latest json files are available here: https://github.com/guneetnarula/indian-district-boundaries/tree/master/topojson/state-wise 

In my last PR, jsons for Andhra Pradesh, J&K and Ladakh were integrated. So use the others for now. I'll upload the remaining within a day.

sivaanil:
@guneetnarula Little curious May I know from where can you get these files ?

guneetnarula:
@shuklaayush @JunaidBabu @sudevschiz 
Almost all of the topojsons are available now in the repo link above. Will finalize Tamil Nadu and the remaining states of northeast once we have the latest available.

@sivaanil - these are community led efforts that are more or less manual in nature. See projects.datameet.org/maps/ for more (although we still have to update the datameet repositories with the latest)

sivaanil:
Thank you for the info @guneetnarula

limshark:
I noticed Gwalior District and Datiya District map in Madhya Pradesh state are not showing correctly, Not sure if I need to file a seperate issue or it is taken care by Json file udpate. the official map is hosted at  http://geoportal.mp.gov.in/DistrictPortal.aspx   . Both district you will find in north of Madhya Pradesh map.   The Covid19india site is showing incorrectly.

limshark:
Both are at the north of Madhya Pradesh.

psunthar:
Several maps are not scaled and translated with the actual GIS boundaries.
Tamil Nadu, for instance.  I tried playing around with the transform.scale and transform.translate values, but it is very difficult to determine the correct position by trial and error.

Any clues how to do this easily?

shuklaayush:
@psunthar d3 has [fitSize/fitExtent](https://github.com/d3/d3-geo#projection_fitExtent) methods for this purpose

guneetnarula:
[...]

PR #420, opened 2020-03-31
Update Bellary to Ballari

salmandaw:
Name change to reflect data


guneetnarula:
Other districts also named incorrectly right now. Related to #413

guneetnarula:
This will be resolved by #521

PR #421, opened 2020-03-31
added space between home-left and home-right for better user visibility

Ankit895:
From 
Screenshot from 2020-03-31 17-41-51

to 
Screenshot from 2020-03-31 17-41-45

for better user visibility so that user can easily read.

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -148,10 +148,14 @@ $purple-light-opaque: #e3e2f3;
 @mixin levelColor($color) {
   .is-#{$color} {
 
-    h1 { color: $color; }
+    h1 {
+      color: $color;
+    }
 
     h4,
-    h5 { color: #{$color}-mid; }
+    h5 {
+      color: #{$color}-mid;
+    }
   }
 }
 
@@ -162,7 +166,8 @@ body {
   -webkit-touch-callout: none;
 }
 
-h1, {
+h1,
+  {
   font-family: 'archia';
   font-size: 24px !important;
   font-weight: 900;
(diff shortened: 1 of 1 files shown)


vramana:
@Ankit895 There are a lot of formatting changes can you undo them, and put them in a different commit or PR. @jeremyphilemon which is better?

Ankit895:
I will update the pr

Ankit895:
@vramana I have update the code, you can review and merge it

vramana:
@Ankit895 It seems it just fixes the problem for your desktop screen. If you shrink the browser window a little more the problem will appear again. I don't think your fix is the correct one. What do you think?

1230 px viewport width

image

Ankit895:
@vramana Thanks bro for bringing this to notice, updated the PR. Please let me know if you find any other issue

src/App.scss
```diff
@@ -223,14 +223,18 @@ h6 {
   flex-direction: row;
   justify-content: space-around;
   flex-wrap: wrap;
-  margin-left: 10rem;
-  margin-right: 10rem;
+  margin-left: 8rem;
+  margin-right: 8rem;
```
vramana, line 227: Is this change even necessary with the margin-left ?

Ankit895: Yes, we need to decrease the margin from home else how would we give margin-left as there is no space to accommodate margin-left

vramana: Did you check it? From what I see it's not necessary.

Ankit895: Yes, I have checked it

Ankit895:
@jeremyphilemon can you please review this?

jeremyphilemon:
Thanks for the PR! I went ahead and patched it using a different strategy since settings margins wouldn't be of much help for devices having larger resolutions anyway.

PR #432, opened 2020-03-31
Feature: Localization support

jeremyphilemon:
* Localization support
* Telegram translation fix
* District wise translations
* Hover over district issue
* Translations cleanup

Feature added by: @harshzalavadiya 

Co-authored-by: Jeremy <jeremyphilemon@outlook.com>


jeremyphilemon:
@harshzalavadiya Hey, so this feature is still in staging and we're looking forward to merge soon! 
There seems to be a few bugs like when you click on Arunachal Pradesh it seems to be crashing, similarly with Meghalaya too. 

I'm not familiar with with i18n library so I was stumped to figure out how to fix it myself, would appreciate it if you could fix that!

Also we need to support these languages on the first push:
- [ ] English
- [ ] Hindi
- [ ] Telugu
- [ ] Kannada
- [ ] Gujarati
- [ ] Marathi
- [ ] Tamil
- [ ] Bangla
- [ ] Odiya
- [ ] Punjabi
- [ ] Malayalam

I think there are json files for languages not mentioned above, feel free to keep them also! More the languages, the better!

You don't have to work on all the translations, it's under stable how much an arduous task that is!
It'd be great if you could just add those locale templates and we can crowdsource for fixes and translations after that!

Let me know what you think :) Really awesome work btw!

Heads up: This is a separate branch now, so your PRs will have to build upon this!

harshzalavadiya:
sure @jeremyphilemon I'll be happy to fix

src/App.js
```diff
@@ -15,11 +15,16 @@ import Links from './components/links';
...
 import Banner from './components/banner';
+import Popup from './components/popup';
 /* import PatientDB from './components/patientdb';*/
 
+import {useTranslation} from 'react-i18next';
+
 const history = require('history').createBrowserHistory;
 
 function App() {
+  const {t} = useTranslation();
```
puru0019: const { translate } = useTransaltion();

`translate` would be more meaningful than just `t`

harshzalavadiya: @puru0019

Hook is from react-i18next and we can't change it directly https://react.i18next.com/latest/usetranslation-hook

KrishnaPravin: if `translate` is a must, then this might help
```js
const { t: translate } = useTransaltion();
```

harshzalavadiya: IMHO, @KrishnaPravin, I know aliasing is possible but doing it everywhere, or wrapping it in another hook isn't a clean approach.

they are naming function as 't' because it's less distractive provides better DX, ctrl clicking on t goes to useTranslate that anyway says that this is a translate function plus most react translation libraries do it same way.

https://www.npmjs.com/package/react-translate
https://react.i18next.com/latest/usetranslation-hook

KrishnaPravin: understood👍

harshzalavadiya:
@jeremyphilemon I guess this branch is far outdated to be merged with master,

Humble advise: next time if localisation happens it should be merged with master in day or two since it affects almost all files, don't wait for all translations to be perfect Just merge it once and then keep on improving it. Otherwise merge conflicts will keep on happening and it will end up in similar situation

Anyways it was a nice learning experience for me keep up the good work

SivakumarNuthalapati:
Please Check Below Error...

Screenshot 2020-04-14 at 7 05 04 PM

jeremyphilemon:
Closing in favor of #1862 :)

PR #423, opened 2020-03-31
added dark mode in App.js

devmanojsoni:
Before:
before_1
before_2
before_3

After:
after_1
after_2
after_3

PR #116, opened 2020-03-25
feature: Darkmode

abhishekkhandait:
Added darkmode for visual ergonomics. 

Support for themes(light and dark) is also added. The current color scheme is used as 'light' theme and a new color scheme for dark mode is added.

PR #433, opened 2020-04-01
added dark mode feature with floating button

devmanojsoni:
before dark mode:
1before_1

after dark mode:
1after_1


shuklaayush:
@devmanojsoni Thanks, but we need more discussion on this before we entertain PRs. Refer to #116 and #423. Closing

PR #505, opened 2020-04-01
Adding info of most recent case

### LINKED ISSUE #443: Last case reported from state
It would be easier to know if along with last updated, there is a way to check last case reported from state.

Ex:
LAST UPDATED 4 MINUTES AGO 01 APR, 14:37:25 IST
LAST CASE REPORTED FROM TAMILNADU,

AkhileshK8y:
**Description of PR**
Home page shows last updated time. It doesn't tell anything about
the state from which most recent case was found. This PR tells
about the most recently affect state along with the last updated
time.

Most recently affected  state is calculated based on the last
updated time of each of the state in data.json API.
**Type of PR**
- [ ] Bugfix
- [*] New feature

**Relevant Issues**  
Fixes #443 

**Checklist**
- [ ] Properly formatted
- [*] Tested on desktop
- [*] Tested on phone

**Screenshots**
Screenshot 2020-04-02 at 3 39 31 AM

Signed-off-by: Kommireddy Akhilesh<akhileshkommireddy2412@gmail.com>

### DIFF
--- a/src/components/home.js
+++ b/src/components/home.js
@@ -8,12 +8,26 @@ import MapExplorer from './mapexplorer';
 import TimeSeries from './timeseries';
 import Minigraph from './minigraph';
 
+function getLatestState(stateWiseData) {
+  return stateWiseData
+    .map(function (e) {
+      if (e.state !== 'Total') {
+        return [formatDate(e.lastupdatedtime), e.state];
+      } else {
+        return ['0000-00-00T00:00', e.state];
+      }
+    })
+    .sort()
+    .reverse()[0][1];
+}
+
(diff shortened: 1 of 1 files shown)


raguljr:
I think, Last case reported from or Most recent case reported from will be more meaningful

puru0019 (approved):
Did you consider of using styled-components? So that you can eliminate repeating same styles in multiple places. I can understand it is not the correct time to look at the quality of code but when ever you have time please implement it

jeremyphilemon:
Hey thanks for the PR! We used a different strategy to solve this using the updates section :)

PR #514, opened 2020-04-02
Added support for showing latest reported cases

jeremyphilemon:
🤷‍♂ 

Screen Shot 2020-04-03 at 5 17 05 PM


src/components/patients.js
```diff
@@ -53,35 +54,40 @@ function Patients(props) {
...
-          } else {
-            return null;
-          }
-        })}
+        {Object.keys(logs)
+          .slice(props.summary ? -1 : 0)
+          .map((day, index) => {
+            if (day !== 'Invalid Date') {
+              return (
+                <React.Fragment>
```
vramana: `key` prop is missing

vramana (changes requested):

src/components/patients.js
jeremyphilemon: Haha fixed :)

saquibkhan:
@jeremyphilemon Please check console. There is an error.

```TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at X (downloadblock.js:10)
    at Qo (react-dom.production.min.js:153)
    at vu (react-dom.production.min.js:261)
    at lc (react-dom.production.min.js:246)
    at uc (react-dom.production.min.js:246)
    at Zu (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at e.unstable_runWithPriority (scheduler.production.min.js:19)
    at qi (react-dom.production.min.js:122)```

PR #521, opened 2020-04-02
Fix: District Maps

jeremyphilemon:
Thanks to @guneetnarula 🎉 

Repo to look out for: https://github.com/guneetnarula/indian-district-boundaries/tree/master/topojson/state-wise


vishalkumar96:
Code is not working on prd kindly check it

Issue #549, opened 2020-04-02
Contribution from state or district is showing 0.00% while it should be 0%


ashishdawale20:
**Describe the bug**
Contribution from state or district is showing 0.00% while it should be 0%

**To Reproduce**
Hover on state Meghalaya, it is showing "0.00% from Meghalaya"  

**Expected behavior**
It should show "0% from Meghalaya" on hover

**Screenshots**
![image](https://user-images.githubusercontent.com/43884208/78274206-78ddc780-752d-11ea-97e1-842cefdc5666.png)

**Desktop (please complete the following information):**

- OS: Windows
- Browser --chrome

PR #550, opened 2020-04-02
Fix to the #549

ashishdawale20:
**This is small fix to the issue #549**

**Type of PR**

- [ #549] Bugfix

**Checklist**

- [ ] Properly formatted
- [ ] Tested on desktop

**Screenshots**
image


shuklaayush:
Thanks @ashishdawale20! But this PR needs more work. As is, it won't fix the likes of 2.00% or 0.20%. A better approach would be to trim trailing zeros after the decimal.

ashishdawale20:
@shuklaayush - Done. trimmed trailing zeros

image

image


image

ashishdawale20:
@shuklaayush  - Does it look good?

PR #593, opened 2020-04-03
Fix: Implement sanity check

### LINKED ISSUE #544: Add sanity check to only plot spread trend graphs till yesterday
**Describe the bug**
The Cases_Time_Series sheet is vulnerable to extra rows of dates being added. This will cause an empty value to be added. Things like below could happen

**To Reproduce**
1. Add an extra date in Cases_Time_Series sheet without any valid Confirmed/Recovered/Deceased values and wait for <9 minutes

**Expected behavior**
Even if empty dates are present, they shouldn't be plotted. Only rows with a valid value should be plotted.

**Screenshots**
image

**Additional context**
We had decided earlier to only show spread trend till yesterday so that there are no 0 values at the edge of the graph,

vedadeepta:
**Description of PR**
Implement sanity check to case_time_series data to check if all the fields & date are present and all the dates <= yesterday

**Type of PR**

- [x] Bugfix
- [ ] New feature

**Relevant Issues**  
Fixes #544 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone


vedadeepta:
@jeremyphilemon this is not a WIP. Its done from my side unless any specific changes you want me  to make ? if everything looks good to you then i can squash to a single commit or GitHub has sqash & merge feature i think - whichever you prefer :)

jeremyphilemon:
This is awesome haha! Thank you for this :)

PR #598, opened 2020-04-03
Fixed modal overflow scrollbar and background scroll

damnitrahul:
Changed `overflow-y:scroll` to `overflow-y:auto` to remove scrollbar when not needed

**Screenshots**

**Before** 
image
**After**
image


damnitrahul:
Fixed background scroll on model open - fixed #599

PR #626, opened 2020-04-04
Made the Helpful links open in new tab - links.js

denvamsi:
**Description of PR**
I would totally recommend this, Because when people wants to look for other sources they don't need to go away from the website so it keeps in engaging :)

**Type of PR**

- [✔] Enhancement

**Checklist**

- [✔ ] Compiles and passes lint tests
- [ ✔] Properly formatted
- [✔ ] Tested on desktop
- [ ✔] Tested on phone


jeremyphilemon:
Patched :)

PR #606, opened 2020-04-03
fixed modal fully

Binu42:
**Description of PR**
stopped scrolling of page when modal is open and close the modal from anywhere outside modal on click.
**Type of PR**

- [x] Bugfix
- [ ] New featur

- [x] Compiles and passes lint tests
- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
##### Before
gify

<br>

##### After
After

PR #661, opened 2020-04-04
Enhancements for DB

jeremyphilemon:


denvamsi:
Hey this looks good, But you can use this #606 and there mr Binu42 has added an feature to close box via clicking on out of the box, This makes the user experience good :)

denvamsi:
But still it has some design issues :)

dilawar:
Great! I was using python+lxml to get the table out of the published google sheet.

Is there a link to the CSV/db file which one can use in analysis pipeline? Sorry but I couldn't find one on the main site.

PR #668, opened 2020-04-05
Reveal states with animation

rbharadbb:
**Description of PR**

Added flip animation to each district row when a state has been revealed
Also fixes the unknown district background color for odd total of districts

**Type of PR**

- [X] Bug fix
- [X] New feature

**Checklist**

- [X] Compiles and passes lint tests
- [X] Properly formatted
- [X] Tested on desktop
- [X] Tested on phone

**Screenshots** (gif does not show convincing animation)

state-reveal-after


src/App.scss
```diff
@@ -164,6 +164,89 @@ $pink-mid: #ffa8cb99;
...
+  70: 3550ms,
+  71: 3600ms,
+  72: 3650ms,
+  73: 3700ms,
+  74: 3750ms,
+  75: 3800ms,
+  76: 3850ms,
+  77: 3900ms,
+  78: 3950ms,
+  79: 4000ms
```
rbharadbb, line 247: At the moment Uttar Pradesh has the highest number of districts (75).
And here we have handled animation for 80 districts.

bee-rickey:
What's the use case for this? As I see it, this does not help the cause in any way.

bee-rickey:
Closing this PR for now.

rbharadbb:
> What's the use case for this? As I see it, this does not help the cause in any way.

@bee-rickey Thank you for looking into it. This PR fixes one bug - Background color of unknown district is always white. Where it should be grey in case the last but one row is white.

Before:
Screenshot 2020-04-11 at 8 09 38 PM

After:
Screenshot 2020-04-11 at 8 10 25 PM

PR #673, opened 2020-04-05
Spread trends filtered by state 

### LINKED ISSUE #80: Statewise trend graph

Now that half the states /UTs(13) have double-digit cases, it may be better time to add statewide trends in `SPREAD TRENDS` as cases are doubling every day in few states.

A user can select the state from the list of states, the trend can be displayed rather than having a single map for each state.  By default, the highest state graph will be displayed.

**Additional context**
Along with issue #12, this may help in tracking statewide spreads.

sksavant:
**Description of PR**
Use raw_data.json to dynamically generate timeseries for confirmed cases, filtered by state.

**Type of PR**

- [ ] ~Bugfix~
- [x] New feature

**Relevant Issues**  
Fixes #80 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone


jeremyphilemon (changes requested):
Hey @sksavant, thank you for this PR. I had already seen your statewise graph posted earlier in the issue itself, so I'll have this assigned to you since I've seen a working proof of concept.

So here are the changes I'd like you to make. So we won't be using `data.json` for getting values, the file is pretty big and not meant to be for the website. However, we just recently updated the api to allow for statewise timeseries daily values here https://api.covid19india.org/states_daily.json.

So how this would work is when the rows are hovered, the current graphs would switch to that state. But I can understand if that would be a little complicated, so you can go ahead and make the three graphs as a separate section for now :)

sksavant:
@jeremyphilemon : I have made the requested changes and resolved merge issues with master branch. Let me know if this is good.

jeremyphilemon:
@sksavant thank you for this PR :)

We pushed new features to the timeseries graph so there were some conflicts, but I'll fix them you don't have to worry! Thanks again :D

PR #697, opened 2020-04-05
Region Highlight: Hovered region shold stay highlighted untill next h…

### LINKED ISSUE #695: Region is not highlighted even though it is showing the statistics about the same region
**Hovered region should remain highlighted until the next hover**
Users hover on a region and get some statistics. Later they move the pointer(mouse) away from the map. It will show data about the same region(state/district) but the region will not remain highlighted.

**To Reproduce**

1. Hover on a region(state/district)
_The region will be highlighted and the Statistics about that region will be displayed_
2. Move the pointer away from the region
3. See error
The region will not stay highlighted even though statistics displaying related to it.

**Expected behavior**
It should remain highlighted until the next Hover(highlight).

**Screenshots**
Screenshot from 2020-04-05 18-34-35

- OS: All Os

sumukhah:
…over.

Hovering on a region highlights the region. On mouse leave, this won't stay
highlighted even though the statistics about the region remains the same.

So highlight the region until statistics remain the same.

- [x] Bugfix

**Relevant Issues**  
Fixes #695

**Screenshots**
Screenshot from 2020-04-05 18-39-08

### DIFF
--- a/src/components/choropleth.js
+++ b/src/components/choropleth.js
@@ -82,16 +82,17 @@ function ChoroplethMap({
         .on('mouseover', (d) => {
           handleMouseover(d.properties[propertyField]);
           const target = d3.event.target;
+          d3.selectAll('.map-hover').attr('class', 'path-refion map-default');
           d3.select(target.parentNode.appendChild(target)).attr(
             'class',
             'map-hover'
           );
         })
-        .on('mouseleave', (d) => {
-          const target = d3.event.target;
-          d3.select(target).attr('class', 'path-region map-default');
-          if (onceTouchedRegion === d) onceTouchedRegion = null;
-        })
+        // .on('mouseleave', (d) => {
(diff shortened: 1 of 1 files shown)


Binu42:
i don't think it's yet solved properly. for eg:- you can highlight two state simultaneously.
Screenshot from 2020-04-06 00-03-44

sumukhah:
Hi @Binu42,
Thanks for the review 
Surprisingly it works in google chrome and firefox, 

chrome
firefox

can you provide some browser specifications?

Binu42:
it's not about browser. hover on map and then move out to table and hover on any state.

shuklaayush:
@sumukhah @Binu42 Thanks for your help, but I would suggest not putting time into this. We know what changes we have to make. It's just that right now, we're not confident if this is better UX.

PR #699, opened 2020-04-05
Adding support to show testing stats

jeremyphilemon:


denvamsi:
Hey there, That looks super cool.
image
But here seems to some issue :)

sibyv:
This Covid-19 tests tracker site https://www.finddx.org/covid-19/test-tracker/ uses this Indian Council of Medical Research site as the source e.g.,https://icmr.nic.in/sites/default/files/press_realease_files/PressRelease_09Mar2020_4PM_IST.pdf  e.g.,https://icmr.nic.in/sites/default/files/whats_new/ICMR_testing_update_05April_9PM_IST.pdf. There seems to be a mismatch using the sources used in this pull request, as the ICMR reported daily test numbers seem to be much higher (*cumulative 89,534 as on 5/4/20 vs. ~16K). I can investigate further for additional sources if you think this is worth incorporating as a source for test numbers if not included already.

anamritraj:
There is a UX issue here as well. 

I'm on my desktop and lets say I want to check the source of testing for Madhya Pradesh. Since this is almost at the center of the map, when I try to move towards the testing source link, which is at the top right corner, my cursor is bound to get over other states and the testing data changes. 

This is more clear in the GIF below. Can't think of any possible workarounds for this rn.

covid19 test

src/components/mapexplorer.js
```diff
@@ -350,8 +365,20 @@ export default function ({states, stateDistrictWiseData, regionHighlighted}) {
...
 
+  useEffect(() => {
+    let found = false;
+    stateTestData.forEach((testObj, index) => {
+      if (testObj.state === panelRegion.name) {
+        found = true;
+        setTestObj(testObj);
+      }
+    });
+    if (!found) setTestObj({});
```
bee-rickey, line 379: This line is causing an infinite call to render(). I guess it doesn't understand that state hasn't changed with {}? Change it to setTestObj(null). Should work.

bee-rickey: Or it might be because you've mentioned testObj in the dependency array? One of the two is the reason.

jayantanth:
Thanks for adding testing data

PR #734, opened 2020-04-06
Feature/resourcetab

### LINKED ISSUE #676: Add page with display and utilize the Crowdsourced Resource List
Team has collected a vast number of resources that would be helpful for the general public here : 
https://bit.ly/covid19resourcelist

This need to be displayed with useful filtering capability in a separate tab in the main website.

1. The sheet needs to be converted into an API endpoint (https://github.com/covid19india/api/issues/73)
Update : This is now available in the endpoint https://api.covid19india.org/resources/resources.json

2. Separate tab called "Resource List" to be added to the main website.

3. Provide a filter-view for user to select the state and/or district. 
4. Provide an option fo the user to select the category for which they need help. Simple icons might be better?

5. Filter the right information from JSON above and show it in a card format to the usre.
6. Necessary disclaimers about the usage of the data needs to be added
7. An option to Report Error/Add new resources to be added.

Please try to use the same color scheme and design paradigm as of the main website.

Currently the resource list is present as a link in "Helpful Links"

sama007:
**Description of PR**
In response to the issue **Add page with display and utilize the Crowdsourced Resource List #676**:

A new tab called Helpful Resources has been developed which lets user to filter the all available resources (testing labs, food distribution organisations etc.). The user can filter by State, City and category of the resources.

**Type of PR**

- [ ] Bugfix
- [x] New feature

**Relevant Issues**  
Fixes #676
Resourcetab2
Resourcetab2_pagination

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**


mayankbhagya:
Hey, shouldn't this tab be called 'organisational support' or simply 'organisations'?
Resources is usually used for beds, ventilators, kits, masks etc.

sama007:
Okay. I will simply change the tab name to Supportive Organisations. Does it sound okay? Other feedbacks?

sama007:
- Changed tab name to Essentials. 
- Added mobile view support

sama007:
Attaching screenshots from mobile view
mobile_screenshot 2
mobile_screenshot 1

sudevschiz:
Nice!
Please fix the conflicts. Will assign @jeremyphilemon  for review

JunaidBabu:
Is it possible to add a free-form search field also?

For example, have a search box and the moment I start typing list up everything that matches my input. Since the data is already fetched in one go, there is no network load here.

The reason I'm suggesting this is, if you select Kerala for example, the next box has too many categories and most of them are nearby places. it's difficult to select each and every place and then click search which returns one or two results only

sama007:
Yes. Its possible and I will do it asap. The workload in offc is heavy. So u can expect this by today mid eve.

sudevschiz:
@sama007  Thanks for the PR and hard work. 

Let me report some non-critical bugs. 
1. The location of filters keep jumping based on the length of the list it populates.
2. This is probably a data-side error : Is it possible to standardize and clean the "City" field further? Also should it be mandatory? If people can want to see all the items in a category without selecting a city, that could be allowed isn't it?
3. I see a box after Page 1 of 1? Is that necessary?
4. (Possible future feature) Are the categories pre-defined and limited? In that case, should it be icons instead of dropdown?

Please do have a look at these. 
let me push this to master :)

shuklaayush:
Hi @sama007, can you fix the warning and remove the logs? Also, is the dependency on bootstrap necessary?

sama007:
@sudevschiz 
- I deliberately kept the dropdown size dynamic. Because some options are so long that it cant be accomodated wholly. A fixed large width wasnt looking good in mobile view when the option is a small string. I will change this if required.
- Cleaning up the city field is hard as I cant predict the pattern beforehand. But you are right I should have kept an option called "All cities". My bad. I will fix it. 
- The 'Page 1 of 1(n)' tells us the total number of pages present e.g. Pge 1 of 7. There is no other option to let the user know that there are total 7 pages.
- Categories are neither predifined nor limited. If someone adds a new category to the DB, it will be dynamically added (given that the selected state and city has that category)

Also, as @JunaidBabu mentioned, I am trying to put a global search box for the table. It will help the user to search any keyword. The work is in progress.

@shuklaayush

- Yeah, I should have commented out the console.log lines. Will fix.
- No, bootstrap isn't used. I forgot to remove it from package.json. Sorry about that. Will fix.

JunaidBabu:
Categories selector works only if City is selected. We should make it available for "All Cities" also.
_Reminding the global search too!!_

sama007:
@JunaidBabu 
Ok. Need one clarification. Suppose the user apply filters and click search button. The filtered result arrives. Now, he searches for some keyword in the search box. Should the keyword search be applied globally on all the data or only on the filtered result?

nikita-11:
Hi Junaid. For the initial draft and design been decided for the page, we didn't go ahead with adding too many features. There is a roadmap thought of for next iterations. We will still need some Conversation over "Global Search" bar that you are suggesting. Also the All cities, part has been added for now, but as discussed in the group internally that was supposed to be avoided as we didn't want to display overwhelming list of results all at once.

PR #796, opened 2020-04-08
state timeseries

### LINKED ISSUE #564: State specific Spread Trends when a state is selected
No

When a state is selected on the India map, the corresponding spread trend plots of that state can replace the overall trend plots. That will allow us to see trend of individual state, which should be different for different states, providing additional insights.

KrishnaPravin:
**Statewise timeseries and nav buttons for mobile**

**Type of PR**

- [ ] Bugfix
- [x] New feature

**Relevant Issues**  
Fixes #564 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [ ] Tested on phone

**Screenshots**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1456,6 +1456,11 @@
         "@types/yargs": "^13.0.0"
       }
     },
+    "@juggle/resize-observer": {
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/@juggle/resize-observer/-/resize-observer-3.1.3.tgz",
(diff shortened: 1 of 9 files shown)


shuklaayush:
@KrishnaPravin Nice work! Looks good, except for the constant wiggling on the charts due to the animations. Also, going back to the pan-India charts is a bit unintuitive. I think a better UX would be to only switch the charts when a row is expanded/state is clicked.

@jeremyphilemon Thoughts?

Issue #777, opened 2020-04-07
Update section better view and enhancements


userhasnoname0:
**Is your feature request related to a problem? Please describe.**
Update should come on top for better readablity.

**Describe the solution you'd like**
![image](https://user-images.githubusercontent.com/62875345/78698754-6c3be380-7920-11ea-8fa9-ec8b04c7a48c.png)

userhasnoname0:
For reference

Added below in the beginning of function Home
`const [showUpdate,setShowUpdate] = useState(false);`

Added at the very beginning of <div className="home-right">
`<div className="update-block">
        
        {showUpdate && (
          <React.Fragment>
            
            <div
            className="updates-header fadeInUp"
            style={{animationDelay: '1.5s'}}
          >
            <h1>Updates</h1>
            <h2>{format(new Date(), 'd MMM')}</h2>
          </div>
          <div className="updates fadeInUp" style={{animationDelay: '1.7s'}}>
            {activityLog
              .slice(-3)
              .reverse()
              .map(function (activity, index) {
                return (
                  <div key={index} className="update">
                    <h5>
                      {formatDistance(
                        new Date(activity.timestamp * 1000),
                        new Date()
                      ) + ' Ago'}
                    </h5>
                    <h4>{activity.update}</h4>
                  </div>
                );
              })}
          </div>
          </React.Fragment>
        )}
        </div>`

userhasnoname0:
Updated with beacon and removed date and time for individual update log
![image](https://user-images.githubusercontent.com/62875345/78764182-cd58cb00-79a3-11ea-94e4-0cfcbc8bf38e.png)

userhasnoname0:
Multiple update logs are seprated in json using '\n' , which causing a mess message in update 

Issue:

![image](https://user-images.githubusercontent.com/62875345/78770170-d9e12180-79ab-11ea-8daa-746eb594458b.png)

Could be resolved by separating '<br/> '  atleast. like ,

![image](https://user-images.githubusercontent.com/62875345/78773610-1cf1c380-79b1-11ea-9d81-c1b8ed2d8235.png)

shuklaayush:
Hi @userhasnoname0 , looks like you resolved the problem with git. Regarding this issue, my opinion is that the map should have a higher priority than the Updates section. If we want to show the latest update, we can include it in the banner, and maybe a click on it would scroll to the update section. I don't see the point in moving the whole update section to the top.

userhasnoname0:
[...]

PR #799, opened 2020-04-08
Update section in home-right with beacon and time changes

userhasnoname0:
for #777


jeremyphilemon:
Hey thanks for the PR! Unfortunately, we had to use a different strategy to show the updates when we pushed the redesign today. However, thank you so much for your patience - feel free to open PRs against available issues, we'll definitely keep an eye out for them this time :)

PR #834, opened 2020-04-09
Fix panels not updating

shuklaayush:
Fix panels not updating on hovering table


src/components/mapexplorer.js
```diff
@@ -308,7 +308,7 @@ export default function ({
       const newMap = mapMeta['India'];
       setCurrentMap(newMap);
       const region = getRegionFromState(regionHighlighted.state);
-      setCurrentHoveredRegion(region);
+      setHoveredRegion(region.name, newMap);
```
jeremyphilemon, line 311: Ah nice 😂

shuklaayush: You gotta do what you gotta do

PR #835, opened 2020-04-09
Added state search feature

burlarajesh:
**Description of PR**

**Type of PR**

- [ ] Bugfix
- [x] New feature

**Relevant Issues**  
Fixes #...

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

image


shuklaayush:
@burlarajesh Woah, nice :+1:. Looks great apart from the search query font. Also, do you think we can add functionality to search the districts within this as well?

burlarajesh:
> @burlarajesh Woah, nice 👍. Looks great apart from the search query font. Also, do you think we can add functionality to search the districts within this as well?

Thanks! :) Sure, Let me fix the font and add district filter as well!

burlarajesh:
> > @burlarajesh Woah, nice 👍. Looks great apart from the search query font. Also, do you think we can add functionality to search the districts within this as well?
> 
> Thanks! :) Sure, Let me fix the font and add district filter as well!

@shuklaayush Hey Ayuash, I have fixed search font. However, for the district wise search, I have done a PoC in my local but I suspect having cumulative search along with districts might create confusion to the end-user.  In practice, let's say the user enters the term 'Delhi' in the search box, the table filters both State and District with 'Delhi' which is irrelevant since Delhi state should return all districts in 'Delhi' state but it wouldn't as per the implementation. Attaching the screenshot below for your reference.

image

Please let me know your thoughts on this. For the time being, I would suggest we can let go of this feature and work on district wise feature filters separately.

JunaidBabu:
Thanks for the PR. 
Feature is nice.. but not very helpful here. Closing this for now.
We might need this search feature in the other tabs like patientdb or essentials. Please have a look into what can be done there.

PR #850, opened 2020-04-09
fix stutter in graph on state change from table

KrishnaPravin:
**Type of PR**

- [x] Bugfix
- [ ] New feature

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

### DIFF
--- a/src/components/home.js
+++ b/src/components/home.js
@@ -74,8 +74,8 @@ function Home(props) {
     setActiveStateCode(state ? state.statecode : 'TT');
   };
 
-  const onMapHighlightChange = ({statecode, name}) => {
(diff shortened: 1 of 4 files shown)


shuklaayush:
Was it because of the 'touchstart' bug?

KrishnaPravin:
> Was it because of the 'touchstart' bug?

no @shuklaayush it was because of a rendering logic. I made an extra check before updating the state in the home component. But that extra check created a new reference of an event handler function every time, which was listed as a dependency of useEffect in the mapexplorer component.

PR #981, opened 2020-04-12
Feature/essentials

sama007:
**Description of PR**

**Type of PR**

- [x] Bugfix
- [ ] New feature

**Relevant Issues**  
Fixes #...

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**


bee-rickey (commented):
Can we have the search button along with the rest of the buttons? Feels unintuitive to have it all by itself.

nikita-11:
Added changes for version 2.0 :

Tech:
1. Remove pagination and keep it infinite scroll. Add got to top floating button on bottom right
2. Links open in New tab. Not losing traffic by redirecting the current page to services page
3. All three filters should be optional. If any services is not available for particular location, show a message
4. Remove "APPLY FILTERS AND HIT SEARCH"
5. Add Disclaimer
6. Add trackers from GA
7. Hyperlink Phone number to open dialer in phone
8. Free search field to search data globally
9. Beautify filters
10. Standardise Mobile and Web
 Header: City and Category
 Body: City, Category, Org, Desc, Phone, and Source in that order
11. Alignment for mobile accordian
12. One accordian at a time, other should auto close
13. Links for submit new entry and feedback
14. Share button.

sama007:
@bee-rickey 
From UX perspective, we thought putting buttons with related functionalities together will be more intuitive. The user will start selecting dropdowns from left and come to the far right end to hit the search button. If we put it on the new line, the user may click the other buttons mistakenly (e.g. Add, Share etc) just to be redirected to new page. That seemed frustrating. So we went ahead with this design.

jeremyphilemon:
In terms of function this is really really good! Unfortunately, I feel it's a little too overkill to include the entire material UI library (react-feather was already being used), so need to re-work on the stylesheet anyway. Hopefully that's for another PR!

Once the conflicts are fixed, we can push this to master! I'm keeping an eye on this PR since a lot of work has been put into this!

public/index.html
```diff
@@ -46,8 +55,13 @@
     <meta property="twitter:image" content="/thumbnail.png">
   </head>
   <body>
+    <!-- Google Tag Manager (noscript) -->
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-KH3S5G4"
```
tannakartikey: @sama007 Don't we have GA already?
https://github.com/covid19india/covid19india-react/blob/8b979ecc855593289bc139cfd2e3c782cdfe6377/public/index.html#L77

src/components/resources.js
```diff
@@ -1,16 +1,75 @@
...
+import FormControl from '@material-ui/core/FormControl';
+import Popover from '@material-ui/core/Popover';
+import Select from '@material-ui/core/Select';
+import {makeStyles} from '@material-ui/core/styles';
+import InputLabel from '@material-ui/core/InputLabel';
+import ErrorOutlineOutlinedIcon from '@material-ui/icons/ErrorOutlineOutlined';
+import Fab from '@material-ui/core/Fab';
+import NavigationOutlinedIcon from '@material-ui/icons/NavigationOutlined';
+
+export const useFormControlStyles = makeStyles((isDesktop) => {
```
tannakartikey, line 13: If we are using the variable in the same file then we don't need to export it.

src/components/resources.js
```diff
@@ -112,22 +200,85 @@ function Resources(props) {
     return Object.keys(resourcedict)
       .sort()
       .map((x) => (
-        <option key={x.id} value={x}>
+        <option
```
tannakartikey, line 203: We are using this same map multiple times in a file. Can we extract it to a separate function?
https://github.com/covid19india/covid19india-react/blob/8b979ecc855593289bc139cfd2e3c782cdfe6377/src/components/resources.js#L203-L212

public/index.html
sama007: Yes. There was already a GA tag was being used. But I guess that was for the entire site. As the Essentials tab has a lot of user interaction(button clicking, accordion expanding, keyword searching etc.), we wanted to keep a separate track of this. So, separate GA tag was generated. This GA tracking used in this page some separate triggers and events to specifically monitor the ux of this page.

sama007:
@jeremyphilemon 
Yes, we could have done this entirely writing custom css. In the first release I tried that also. But the widgets were looking different in different desktop environments. In my Gnome Firefox, the native dropdowns were looking good but in Windows it looked pretty lame. So, to save time and effort to reinvent the wheel, I chose react-material.
Regarding react-feather, it has only icons. I can change material-icons to react-icon easily if its needed.
I am working on the merge conflicts.

src/components/resources.js
```diff
@@ -207,80 +380,495 @@ function Resources(props) {
...
-        <div className="resourcefilters">
-          <div className="resourcefilter">
-            {/* <label for='stateselect1' className='filterlabel'>
-                            State
-                        </label> */}
-            <select
-              id="stateselect1"
-              onChange={changeIndianState}
-              defaultValue=""
+        {!isDesktop && (
```
tannakartikey, line 436: This component is huge, almost 200 lines and the whole component is repeated in an if-else condition. I believe for mobile and for desktop. Can we refactor this code to use only one component and then apply CSS/classes using the condition. It will make the code DRY and more readable.

tannakartikey (changes requested):

src/components/resources.js
tannakartikey: We can write CSS in a separate file and include

public/index.html
tannakartikey: @sama007 I think if the multiple tags are not configured correctly when they are included then it will start sending out incorrect data. Can you please confirm?

src/components/resources.js
sama007: Yes I know the files have gotten huge. After this release, our first priority is cleaning the code. When I started this, I didn't anticipate this tab would grow this huge. There are lots of developers working on this and we didn't get time to cleanly merge the changes. We will refactor this code surely.

tannakartikey (approved):

PR #986, opened 2020-04-12
#355 Comparing data from the API and the public/maps file

tannakartikey:
**Description of PR**

The spec gets the data from https://api.covid19india.org/state_district_wise.json and check if each state and district received from the API exist within the `public/maps` file or not.

We need this spec because each day we are getting new places added to the patients' database and hence the API. If any of the upcoming places is missing in the maps data then this spec will start failing. And it is very likely that the place is missing or spelled wrong. It's hard to get the latest data of India in geo objects. 

I am going to comment on PR explaining the output of each function shortly after creating the PR.

The spec is added here instead of API because the map is rendered from the `public/maps` folder and that data is not available in the API repository.

P.S. Special thanks to @guneetnarula and his friends for real hard work and keep the data updated.
Geo objects repo: https://github.com/guneetnarula/indian-district-boundaries.

Note:
The spec will fail early on the first mismatch. With Jest, for our case, it is not possible to make the test fail after checking everything without storing the response of the API in a file first. It will make the test static and useless because we will have to update the data manually.
Related Jest thread: https://github.com/facebook/jest/issues/1619

**Type of PR**

- [ ] Bugfix
- [ ] New feature

**Relevant Issues**  
#355

**Checklist**

- [ ] Properly formatted
- [ ] Tested on desktop
- [ ] Tested on phone

**Screenshots**

### DIFF
--- a/public/maps/mizoram.json
+++ b/public/maps/mizoram.json
@@ -3345,7 +3345,7 @@
           "arcs": [[0, 1, 2, 3, 4, 5, 6]],
           "type": "Polygon",
           "properties": {
-            "district": "Aizawal",
+            "district": "Aizawl",
             "dt_code": "261",
             "st_nm": "Mizoram",
             "st_code": "15",
--- a/src/__tests__/mapAndApiStateNames.test.js
+++ b/src/__tests__/mapAndApiStateNames.test.js
@@ -1,0 +1,25 @@
+import {
+  getStatesAndDistrictsFromAPI,
+  getStatesAndDistrictsFromMaps,
+} from '../testUtils/index';
+
+describe('Compare the map and the API states and districts', () => {
+  test('for any discrepancies', async () => {
+    const statesAndDistrictsFromAPI = await getStatesAndDistrictsFromAPI();
+    const statesAndDistrictsFromMaps = await getStatesAndDistrictsFromMaps();
+
+    const statesFromAPI = Object.keys(statesAndDistrictsFromAPI);
+    const statesFromMaps = Object.keys(statesAndDistrictsFromMaps);
+
+    expect(statesFromAPI).toContain(...statesFromMaps);
+
+    statesFromAPI.forEach((state) => {
+      const expectedDistricts = statesAndDistrictsFromMaps[state]?.sort();
+      const receivedDistricts = statesAndDistrictsFromAPI[state]?.sort();
+
+      if (expectedDistricts !== undefined && receivedDistricts.length !== 0) {
+        expect(expectedDistricts).toContain(...receivedDistricts);
+      }
+    });
+  });
+});
--- a/src/testUtils/index.js
+++ b/src/testUtils/index.js
@@ -1,0 +1,49 @@
+import axios from 'axios';
+import {promises as fs} from 'fs';
+
+export function removeFileExtension(fileName) {
+  return fileName.substr(0, fileName.lastIndexOf('.'));
+}
+
+function removeUnknown(e) {
+  return e !== 'Unknown';
+}
+
+export async function getStatesAndDistrictsFromAPI() {
+  const stateDistrictWiseResponse = (
+    await axios.get('https://api.covid19india.org/state_district_wise.json')
+  ).data;
+  const states = Object.keys(stateDistrictWiseResponse).filter(removeUnknown);
+  const result = {};
+  states.map((stateName) => {
+    result[stateName] = Object.keys(
+      stateDistrictWiseResponse[stateName]['districtData']
+    ).filter(removeUnknown);
+  });
+  return result;
+}
+
+export async function getStatesAndDistrictsFromMaps() {
+  const dir = await fs.opendir('public/maps');
+  const districts = {};
+  for await (const dirent of dir) {
+    const fileName = dirent.name;
+    const fileNameWithoutExtension = removeFileExtension(fileName);
+
+    if (fileName === 'india.json') continue;
+    const data = JSON.parse(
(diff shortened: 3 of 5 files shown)


public/maps/tamil-nadu.json
```diff
@@ -1556,7 +1556,7 @@
     "translate": [76.233101, 8.077847]
   },
   "objects": {
-    "tamilnadu_district": {
+    "tamil-nadu_district": {
```
tannakartikey: All the files except `tamil-nadu.json` the districts' data under `<file_name>_district` property. This change fixes that.

src/__tests__/mapAndApiStateNames.test.js
```diff
@@ -0,0 +1,25 @@
...
+    const statesFromMaps = Object.keys(statesAndDistrictsFromMaps);
+
+    expect(statesFromAPI).toContain(...statesFromMaps);
+
+    statesFromAPI.forEach((state) => {
+      const expectedDistricts = statesAndDistrictsFromMaps[state]?.sort();
+      const receivedDistricts = statesAndDistrictsFromAPI[state]?.sort();
+
+      if (expectedDistricts !== undefined && receivedDistricts.length !== 0) {
+        expect(expectedDistricts).toContain(...receivedDistricts);
```
tannakartikey, line 21: The data from the API should be available within the data collected from the `public/maps/*` folder.
The maps data can have more states then the API data because we may not have active cases in all the states.

src/testUtils/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+export function removeFileExtension(fileName) {
+  return fileName.substr(0, fileName.lastIndexOf('.'));
+}
+
+function removeUnknown(e) {
+  return e !== 'Unknown';
+}
+
+export async function getStatesAndDistrictsFromAPI() {
```
tannakartikey, line 12: A snippet of the output. An actual API call is made to the API and we get an object with keys set to the name of the state and the districts of the state as values.
```javascript
{
        'Arunachal Pradesh': [ 'Lohit' ],
        'Dadra and Nagar Haveli': [],
        Tripura: [ 'Gomati', 'North Tripura' ]
}
```

src/testUtils/index.js
```diff
@@ -0,0 +1,49 @@
...
+  const result = {};
+  states.map((stateName) => {
+    result[stateName] = Object.keys(
+      stateDistrictWiseResponse[stateName]['districtData']
+    ).filter(removeUnknown);
+  });
+  return result;
+}
+
+export async function getStatesAndDistrictsFromMaps() {
```
tannakartikey, line 26: A snippet of the output. The function traverses through the `public/maps` directory and gets the collection. We receive an object with keys set to the name of the state and the districts of the state as values.
```
{
        Puducherry: [ 'Puducherry', 'Karaikal', 'Mahe', 'Yanam' ],
        'Andhra Pradesh': [
          'Srikakulam',    'Vizianagaram',
          'Visakhapatnam', 'Krishna',
          'Guntur',        'East Godavari',
          'Prakasam',      'Kurnool',
          'Anantapur',     'S.P.S. Nellore',
          'West Godavari', 'Y.S.R.',
          'Chittoor'
        ],
}
```

public/maps/tamil-nadu.json
shuklaayush: Hey, you'll have to make this change in mapexplorer.js as well. Also, as you mentioned, map geojsons are taken from @guneetnarula 's repository, so we would have to rename them every time they're updated. So please don't make this change unless it's somehow necessary for the tests.

shuklaayush:
Hi @tannakartikey, while this a good test to look for mismatches between the API and the maps, I think this is not what's required right now for CI. What's more important is that we add tests which check whether any inputs (like clicking on the table, map etc.) cause the website to crash.

tannakartikey:
Hello @shuklaayush 
I will try to cover all the scenarios as I move ahead. This spec was written actually a few days back(https://github.com/covid19india/covid19india-react/pull/356) when we had lots of issues coming related to the name of the districts. This is just the same spec in a new PR.

public/maps/tamil-nadu.json
tannakartikey: @shuklaayush I thought about it and then forgot.
Fixed it with https://github.com/covid19india/covid19india-react/pull/986/commits/18978a5c0f90bc3a762aa0e1d56439e3ce2846c3. Thanks

I will submit a PR in @guneetnarula repo as well. Thanks for bringing my attention to that too.

Good catches, man! :+1:

tannakartikey: @shuklaayush over there the name of the file is `tamilnadu.json` and we have `tamil-nadu.json` so do you think we need to rename the file to `tamilnadu.json`?

public/maps/mizoram.json
```diff
@@ -3345,7 +3345,7 @@
           "arcs": [[0, 1, 2, 3, 4, 5, 6]],
           "type": "Polygon",
           "properties": {
-            "district": "Aizawal",
+            "district": "Aizawl",
```
shuklaayush, line 3348: Let's verify this with @guneetnarula as well

tannakartikey: @shuklaayush sure. I have also raised a PR in his repo: https://github.com/guneetnarula/indian-district-boundaries/pull/2

tannakartikey:
@shuklaayush @jeremyphilemon the name of the district is changed on https://github.com/guneetnarula/indian-district-boundaries/pull/2

jeremyphilemon (approved):
Good stuff :)

PR #993, opened 2020-04-12
Fix: Preload fonts to improve performance

abhishekakade:
**Description of PR**
- This will preload the font files instead of waiting for main JS file to parse and then load fonts.
- Since font files or their names are not going to change, it is fine if they don't get `hashed` on each webpack build and so I placed them in `public/fonts` folder.
- Moved font-face declarations of these preloaded fonts to `index.html` to prevent render blocking css (fixed in second commit).
- This is supposed to improve performance and loading but since I could not test the production build version locally or on my own repo with gh-pages (maps and other things were not loading), I only tested the un-minified dev build which will not provide much insight on actual performance of minified production build. So this should be tested in a separate branch first and then merged with the main branch.

**Type of PR**

- [x] Bugfix

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [ ] Tested on phone

**Screenshots**


tannakartikey:
@abhishekakade can't you test on this https://covid19india-react-git-fork-abhishekakade-font-preloading.covid19india.now.sh?

abhishekakade:
@tannakartikey, I did but since both are on different hosting platforms (GitHub Pages and Zeit), I don't know how reliable the test results will be.

jeremyphilemon:
Thanks for the PR :D

PR #997, opened 2020-04-12
Fix: Enable prettier's pre-commit hook on scss files

KrishnaPravin:
**Fix pre-commit hook configuration**
Changes:
- Added the file types to be checked by prettier
- ~~Replace `--write` with `--check` for prettier and removed `--fix`~~.
    - ~~Reason: when the auto-fix is enabled, devs might not be aware of the changes made by these tools and the changes will come along with the commit.~~

**Type of PR**

- [x] Bugfix
- [ ] New feature

**Screenshots**
Screenshot 2020-04-12 at 4 07 56 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,10 @@
     "test": "react-scripts test --testPathIgnorePatterns=src/__tests__/mapAndApiStateNames.test.js",
     "eject": "react-scripts eject",
     "eslint": "eslint .",
-    "eslint:fix": "eslint --fix .",
-    "format": "prettier --write \"**/*.+(js|jsx|json|yml|yaml|css|md|vue)\"",
-    "scsslint": "sass-lint-auto-fix"
+    "eslint:fix": "eslint --fix --max-warnings=0",
+    "format": "prettier --write \"**/*.+(js|jsx|json|yml|yaml|md)\"",
+    "scss-lint": "sass-lint 'src/**/*.scss' -v -f table --max-warnings 0",
+    "scss-lint:fix": "sass-lint-auto-fix 'src/**/*.scss' -d -h -c '.sass-lint-auto-fix.yml'"
   },
   "eslintConfig": {
     "extends": "react-app"
@@ -53,14 +54,15 @@
   },
   "lint-staged": {
     "*.+(js|jsx)": [
-      "eslint --fix --max-warnings=0",
-      "prettier --write"
+      "npm run eslint:fix",
+      "prettier --write \"**/*.+(js|jsx)\""
     ],
-    "*.+(json|yml|yaml|md|vue)": [
-      "prettier --write"
+    "*.+(json|yml|yaml|md)": [
+      "prettier --write \"**/*.+(json|yml|yaml|md)\""
     ],
     "*.+(css|scss)": [
-      "sass-lint-auto-fix"
+      "npm run scss-lint:fix",
+      "npm run scss-lint"
     ]
   },
   "devDependencies": {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,18 +10,24 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: [12.x]
-
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+
+      - name: Cache node modules
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-node-modules
         with:
-          node-version: ${{ matrix.node-version }}
-      - run: npm ci
-      - run: npm test
+          path: ~/.npm # npm cache files are stored in `~/.npm` on Linux/macOS
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install packages
+        run: npm ci
+      - name: Run Test Suite
+        run: npm test
         env:
           CI: true
           FORCE_COLOR: 1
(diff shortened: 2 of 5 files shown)


shuklaayush:
@KrishnaPravin I think @jeremyphilemon deliberately added the '--write' option to make sure that easy fixes like whitespaces, indentations are fixed in place automatically.

KrishnaPravin:
@shuklaayush I totally agree with whitespace and indentation issues. But what about other code changes? I think auto-fix can make changes to the code without the knowledge of the author of the commit. Do you want that to happen?

KrishnaPravin:
Moreover, whitespace, indentation and all other issues can be fixed by running the respective npm scripts. its just `npm run eslint` and `npm run format`

shuklaayush:
> @shuklaayush I totally agree with whitespace and indentation issues. But what about other code changes? I think auto-fix can make changes to the code without the knowledge of the author of the commit. Do you want that to happen?

@KrishnaPravin Let's just wait for @jeremyphilemon before we merge. I'm not an expert on prettier. I assumed that it only does formatting changes. 

> Moreover, whitespace, indentation and all other issues can be fixed by running the respective npm scripts. its just npm run eslint and npm run prettier

This would work. But what I'm afraid is that a lot of people will instead ignore the prettier errors and just make a PR by adding --no-verify.

tannakartikey:
@KrishnaPravin I think we should put back `--write`. The auto-fix will not change code logic or anything. It will just prettify.

KrishnaPravin:
@shuklaayush @tannakartikey reverted back to auto fix

tannakartikey:
@KrishnaPravin started working perfectly for `scss` files.
Would you like to add the lint task in `.github/workflows/ci.yml` for the server-side check?

KrishnaPravin:
@tannakartikey Yes Sure. I will add the check to `ci.yml`

package.json
```diff
@@ -38,9 +38,10 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "eslint": "eslint .",
-    "eslint:fix": "eslint --fix .",
-    "format": "prettier --write \"**/*.+(js|jsx|json|yml|yaml|css|md|vue)\"",
-    "scsslint": "sass-lint-auto-fix"
+    "eslint:fix": "eslint --fix --max-warnings=0",
+    "format": "prettier --write \"**/*.+(js|jsx|json|yml|yaml|md)\"",
+    "scss-lint": "sass-lint 'src/**/*.scss' -v -f table --max-warnings 328",
+    "scss-lint:fix": "sass-lint-auto-fix 'src/**/*.scss' -d -h"
```
tannakartikey: @KrishnaPravin from the docs of `sass-lint-auto-fix`

> By default, sass-lint will look for a .sass-lint.yml or .sasslintrc, or an entry in your package.json. When applying the resolver to a given rule, configuration is parsed from one of the above files..
> https://www.npmjs.com/package/sass-lint-auto-fix#configuring-which-rules-to-run-against

There is no mention of a file named `.sass-lint-auto-fix.yml` to be considered automatically as config. And I did not the see that file provided with `-c` option to the `sass-lint-auto-fix` command.

package.json
```diff
@@ -52,14 +53,15 @@
...
     ],
-    "*.+(json|yml|yaml|md|vue)": [
-      "prettier --write"
+    "*.+(json|yml|yaml|md)": [
+      "prettier --write \"**/*.+(json|yml|yaml|md)\""
     ],
     "*.+(css|scss)": [
-      "sass-lint-auto-fix"
+      "npm run scss-lint:fix",
+      "npm run scss-lint"
```
tannakartikey, line 65: Do we need to run `scss-lint` again after running `scll-lint:fix`? Would not it throw error if it could not fix something?

package.json
```diff
@@ -38,9 +38,10 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject",
     "eslint": "eslint .",
-    "eslint:fix": "eslint --fix .",
-    "format": "prettier --write \"**/*.+(js|jsx|json|yml|yaml|css|md|vue)\"",
-    "scsslint": "sass-lint-auto-fix"
+    "eslint:fix": "eslint --fix --max-warnings=0",
+    "format": "prettier --write \"**/*.+(js|jsx|json|yml|yaml|md)\"",
+    "scss-lint": "sass-lint 'src/**/*.scss' -v -f table --max-warnings 328",
```
tannakartikey: Also, any reason why `--max-warnings` is set to 328? Is that the number of warnings we are getting at the moment?

tannakartikey (changes requested):

package.json
KrishnaPravin: Yes, @jeremyphilemon  is working on reducing the number

package.json
KrishnaPravin: Exactly that is what we want, errors like class-name-format and so many others cannot be fixed automatically. For those cases, we would want to prevent a commit until a manual fix is done.

package.json
KrishnaPravin: We can enable/disable the resolvers using `sass-lint-auto-fix`. For eg: `no-css-comments` this resolver is disabled using `sass-lint-auto-fix.yml` but a warning is enabled in `sass-lint`.

tannakartikey: @KrishnaPravin I got it. But the file `sass-lint-auto-fix.yml` is not given to the command. Should not we have to do something like below?
```
"scss-lint:fix": "sass-lint-auto-fix 'src/**/*.scss' -d -h -c .sass-lint-auto-fix.yml"
```

KrishnaPravin: I will add `-c ` option

KrishnaPravin: I've added it

tannakartikey (approved):

tannakartikey:
Looks good to me except `npm ci` failed again.

KrishnaPravin:
> Looks good to me except `npm ci` failed again.

we can try `npm i` if `npm ci` is not reliable.

tannakartikey:
@KrishnaPravin can you please rebase?

KrishnaPravin:
> @KrishnaPravin can you please rebase?

I guess @jeremyphilemon is working on this

jeremyphilemon (approved):

PR #1012, opened 2020-04-12
select component styling made cleaner

### LINKED ISSUE #1010: Make select options in demographics cleaner
The select options component is compact. It should be more clean and the font should be same as base font (Archia). It looks bad in mobile devices. It should be responsive.

nayakrahul:
The select component in demogarphics has been made cleaner. Styling has been changed.
**Type of PR**

- [ ] Bugfix
- [x] New feature

**Relevant Issues**  
Fixes #1010

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**


amittttt (approved):

nayakrahul:
@amittttt please review and merge !!

src/components/patientdb.js
```diff
@@ -74,8 +96,23 @@ function PatientDB(props) {
       } else if (label === 'detecteddistrict') {
         const city = document.getElementById('city');
         // Hide box
-        if (value === '') city.style.display = 'none';
-        else city.style.display = 'inline';
+        if (value === '') {
+          if (!city.classList.contains('hidden')) {
```
Maskedman99, line 98: combine this to ```if (value === '' && !city.classList.contains('hidden'))``` then ___

src/components/patientdb.js
```diff
@@ -74,8 +96,23 @@ function PatientDB(props) {
...
+            city.classList.add('transition');
+            const width = city.clientWidth;
+            city.classList.add('hidden');
+          }
+        }
+        else {
+          if (city.classList.contains('hidden')) {
```
Maskedman99: __ you can change this to a simple ```else```

src/components/patientdb.js
nayakrahul: **If value is '' and city element is hidden, we don't want any transition or changes to happen.**
If both the conditions are combined, the combined condition will give false on above scenario thus, the next else condition will trigger which we don't want.

src/components/patientdb.js
nayakrahul: I changed it to ```else if```

PR #1082, opened 2020-04-14
Style: Improve upon UI/UX of DeepDive

pastelsky:
**Description of PR**
Improves upon the basic UI/ UX of the deep dive page:

- Use sample size instead of the growing "Awaiting data" count.
- Use subtler colors in charts
- Optimize charts for mobile – packed tighter and less wasted white-space.

**Type of PR**
- [x] Enhancement

**Relevant Issues**  

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

### DIFF
--- a/src/components/Charts/allstates.js
+++ b/src/components/Charts/allstates.js
@@ -1,32 +1,15 @@
+import {defaultOptions, xAxisDefaults, yAxisDefaults} from './chart-defaults';
+
 import {getStateName} from '../../utils/commonfunctions';
 
+import deepmerge from 'deepmerge';
 import moment from 'moment';
 import React from 'react';
-import {Line, defaults} from 'react-chartjs-2';
+import {Line} from 'react-chartjs-2';
+
 function AllStatesChart(props) {
   const dates = [];
   const chartReference = React.createRef();
-
-  defaults.global.elements.line.fill = false;
-
-  defaults.global.tooltips.intersect = false;
-  defaults.global.tooltips.mode = 'nearest';
-  defaults.global.tooltips.position = 'average';
-  defaults.global.tooltips.backgroundColor = 'rgba(255, 255, 255, 0.6)';
-  defaults.global.tooltips.displayColors = false;
-  defaults.global.tooltips.borderColor = '#c62828';
-  defaults.global.tooltips.borderWidth = 1;
-  defaults.global.tooltips.titleFontColor = '#000';
-  defaults.global.tooltips.bodyFontColor = '#000';
-  defaults.global.tooltips.caretPadding = 4;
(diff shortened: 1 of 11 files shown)


vedadeepta:
Facing the same issue with my PR. i think before `npm ci`, `npm install` command is needed.

src/components/deepdive.js
```diff
@@ -43,36 +43,33 @@ function DeepDive(props) {
...
           <AllStatesChart
-            title="States - Total Cases"
+            title="Total Cases by State"
             data={statesTimeSeries}
           />
         </div>
 
         <div className="card fadeInUp" style={{animationDelay: '0.7s'}}>
-          <GenderChart title="Patient Gender" data={rawData} />
+          <GenderChart title="Patient by Gender" data={rawData} />
```
s-naveen: Should be Patients by Gender

src/components/Charts/chart-defaults.js
```diff
@@ -0,0 +1,112 @@
...
+    position: 'average',
+    displayColors: false,
+    borderWidth: 0,
+    titleSpacing: 4,
+    titleMarginBottom: 8,
+    bodySpacing: 8,
+    cornerRadius: 3,
+    xPadding: 10,
+    yPadding: 8,
+    caretSize: 0,
```
s-naveen: I feel tooltip is too big now and some part hides behind bottom of the chart. Need to revisit this.

src/components/Charts/allstates.js
```diff
@@ -94,7 +70,7 @@ function AllStatesChart(props) {
     }
 
     datasets.push({
-      borderWidth: 1.5,
+      borderWidth: 3,
```
s-naveen: it looks good to me but some felt thin lines are better. @sudevschiz @shuklaayush @JunaidBabu thoughts?

pastelsky: I found the updated colour palette for these lines to not work very well with very thin lines. But if you think the thinner ones looked better, I've updated it to `2` now.

src/components/deepdive.js
pastelsky: Good catch, fixed

src/components/Charts/chart-defaults.js
pastelsky: While the current tooltip is concise, I did find issues when trying to actually read the contents with small text on a semi-transparent background.

image
image



I've updated this to use tighter spacing now (still a tad little roomier than exists today)

pastelsky:
@s-naveen Could take another look at this?

jeremyphilemon (approved):
LGTM, thanks for the PR! :D

PR #1155, opened 2020-04-16
Fix: Icon padding in the footer links.

amitmerchant1990:
Hi guys,

Great work on this beautiful portal. However, I found a minor issue in the bottom links where SVG and text are very nearby to each other which doesn't look professional. 

Screenshot from 2020-04-16 14-46-57

So, I've fixed that. Now, it looks like this.

Screenshot from 2020-04-16 14-49-12

Let me know if this makes sense.

Thanks!

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1232,6 +1232,7 @@ h6 {
   svg {
     width: 20px;
     stroke-width: 2px;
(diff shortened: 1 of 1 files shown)


jeremyphilemon:
Looks good, thanks for the quick fix :)

Issue #662, opened 2020-04-05
Please read this before submitting an issue 📮


jeremyphilemon:
First off, thanks for taking the time to report a bug/request a feature :+1:

Two kinds of issues can be raised in this project:

1. Bug report
2. Feature Request

**Please search and see if your issue has already been filed/asked before submitting it. Duplicates will be closed. If your issues do not have the questionnaire filled, it will be closed. If you feel the need to address an already filed issue, please comment in that issue itself.**

## Process Flow

- When a new issue is created, "Status: Review Needed" and "Type: Bug or Type: Enhancement" label is automatically added to it.
- Mods periodically check "Status:Review Needed" and add appropriate Priority and Status labels. If the issue is clear, set the status label as "Label: Available".

<p align="center">
  <img src="https://storage.googleapis.com/jrmyphlmn/images/github/label%20change.png" width="50%">
</p>

- People looking to contribute, filter issues by ["Status:Available"](https://github.com/covid19india/covid19india-react/issues?q=is%3Aissue+is%3Aopen+label%3A%22Status%3A+Available%22) and comment "/claim" on the issue they would like to be assigned to work on.

<p align="center">
<img src="https://storage.googleapis.com/jrmyphlmn/images/github/filter.png" width="50%">
</p>
<h6 align="center">Filter by label "Available"</h6>
<p align="center">
<img src="https://storage.googleapis.com/jrmyphlmn/images/github/claim.png" width="50%">
</p>
<h6 align="center">use /claim to have the issue assigned to you</h6>

- Finally submit the pull request.

sudevschiz:
## **District Name Errors** ##

Before sending feedback or creating an issue that deals with a district name, please note: 
Central and State governments, local authorities and the media are not consistent in how they refer to a district. Registrar General of India, Survey of India, State Govt websites use different spellings or even different names sometimes. In light of this, we are standardizing the district names that we are using in our website and its entire data pipeline. The current list of district names that we are using can be seen in the "District_List" sheet of https://bit.ly/patientdb

If you still would like to report and error, it might be best done through the "Contact Us" form here : https://forms.gle/MLeSrdwRuf8AjFkp9

sudevschiz:
## **Data Errors**

If your issue is regarding an error in the data, it's best reported through this form: https://forms.gle/MLeSrdwRuf8AjFkp9

PR #1208, opened 2020-04-17
FIXED - District name typos for Jammu & Kashmir

ghost:
**Description of PR**
Fixed the names of 5 districts of Jammu & Kashmir which were spelled incorrectly.

**District Names Fixed**
Bandipore => Bandipora,
Baramula => Baramulla,
Badgam => Budgam,
Punch => Poonch,
Shupiyan => Shopian

**Reference**
[Wikipedia](https://en.wikipedia.org/wiki/List_of_districts_of_Jammu_and_Kashmir)

jammu-and-kashmir-districts-wikipedia

### DIFF
--- a/public/maps/jammukashmir.json
+++ b/public/maps/jammukashmir.json
@@ -2305,7 +2305,7 @@
           "type": "Polygon",
           "properties": {
             "dt_code": "009",
-            "district": "Bandipore",
+            "district": "Bandipora",
             "st_nm": "JAMMU & KASHMIR",
             "st_code": "01",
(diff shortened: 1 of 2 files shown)


shuklaayush:
@tauseefsshah Thank for the PR, but please read this as well https://github.com/covid19india/covid19india-react/issues/662#issuecomment-609377552.
Only changing the name in the map files will break data association with the API. For example, you can see that after your changes, Baramulla will no longer be coloured on the map. This requires changes in the API too.

PR #877, opened 2020-04-09
Fix: Matched style of select in demographics to that of homepage

Binu42:
**Description of PR**
* styled dropdown menu across website.
*  prevented entry of invalid page number in input on essential page.

**Type of PR**

- [x] Bugfix
- [ ] New feature

**Relevant Issues**  
Fixes #852, Fixes #747

**Checklist**

- [x] Compiles and passes lint tests
- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
Screenshot from 2020-04-11 09-31-07

Screenshot from 2020-04-10 05-20-25

PR #1228, opened 2020-04-17
fix the dropdown UI bug of PR-877

strikertanmay:
**Description of PR**

**Type of PR**

- [x] Bugfix
- [ ] New feature

**Relevant Issues**  
Fixes #...

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -2493,13 +2493,7 @@ footer {
     text-align: center;
   
     select {
-      appearance: none;
       background-color: #eee;
-      background-image: linear-gradient(45deg, transparent 50%, #808080 50%),
-        linear-gradient(135deg, #808080 50%, transparent 50%);
-      background-position: calc(100% - 13px) 50%, calc(100% - 8px) 50%;
-      background-size: 5px 5px, 5px 5px;
-      background-repeat: no-repeat;
       border: 2px solid #e8e8e9;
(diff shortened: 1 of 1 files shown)


strikertanmay:
@jeremyphilemon This PR will fix out that thing which I discussed on #877

Binu42:
Hi, @strikertanmay this UI is not good because it's occupying half of the screen for filter in mobile view.
Screenshot from 2020-04-17 23-33-20

strikertanmay:
> Hi, @strikertanmay this UI is not good because it's occupying half of the screen for filter in mobile view.
> Screenshot from 2020-04-17 23-33-20

I checked for different media sizes. No such problem occurs.
If this problem occurs it will easily solve by some increment in top padding for that media size category.

strikertanmay:
@Binu42 
Screenshot from 2020-04-17 23-47-57 (1)
I again checked on google chrome and firefox for the same media size mentioned in SS. No such glitch found.
Which browser u are using ?

jeremyphilemon:
I think what @Binu42 is trying to say is that having the filters arranged as a column is occupying a lot of space of the screen.

Would like to have this filed as issue first so that we can think of a potential solution before creating the PR :)

PR #1246, opened 2020-04-18
Refactor: Replace hover events with click on map

### LINKED ISSUE #717: When user mouse hover on State/UT at  the bottom of a list, User need to scroll up screen for more details. So this is not user friendly way!
This is not a user friendly way like When user mouse hover on State/UT at  the bottom of a list I need to scroll up screen for looking the more details. 

If possible you can move a Spreads Trend above and more detail state/UT map section below that and hard code the More Details state section, so we scroll up and down hovering the list of states accordingly it will change the state/UT detail map 

**Additional context**

### LINKED ISSUE #1243: The graph should change data on click, not on hover.
I'm always frustrated when the graph changes automatically on hover. It looks cool but not practical and bad UI choice. To be precise,

- Everytime I want to see "all states", No matter how I go to graphs, I already had hovered over some states whose data is in graph now.  I have to reset it manually.

- If I wanted to change graph to "All states", the drop-down menu goes over the Kerala & Karnataka. If I clicked "All states" The menu disappears and Kerala gets selected. The drop-down menu is also much nearer to map so it become susceptible to other hovering mistakes too.

image

- Let's suppose I wanted to see Maharastras data in the graph. I went to Maharastra in the map, Until I scroll down to the graph, The mouse had hit Kerala or any other south state which already have changed graph data. 

The graph should change data on click, not on hover. Hovering makes it so sensitive to change.  User has to cognitively handle the mouse which is bad UI design. 

Have an "All state button" at top of the graph to reset it. As it is the most seen graph by people, They shouldn't always have to reset it through drop down. Even though It is alternative, we could apply both things rather than just one of them.

shuklaayush:
**Description of PR**
The behaviour is now consistent with touch screens. Single click to select, and double click to expand map.

Also removed back button. Double clicking outside map will now go back to country view.

**PS: This PR is subject to scrutiny if this behaviour does not seem fit for PC.**

**Type of PR**

- [ ] Bugfix
- [x] New feature

**Relevant Issues**  
Fixes #717, fixes #1243 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -445,7 +445,6 @@ h6 {
 
     .header,
     .map-stats,
-    .back-button,
     .last-update,
     .unknown,
     .state-page-button {
@@ -1814,10 +1813,6 @@ table {
   z-index: 10;
 }
 
-.back-button {
-  z-index: 11;
-}
-
 .anchor {
   cursor: pointer;
   position: absolute;
@@ -1913,19 +1908,6 @@ table {
 
     &.deceased {
       stroke: #6c757d;
-    }
-  }
-
-  .back-button {
-    background: $yellow-light;
-    color: $orange;
-    position: absolute;
-    right: 0;
-    top: 2.75rem;
-    transition: all .2s ease-in-out;
-
-    &:hover {
-      background: $yellow-hover;
     }
   }
 
--- a/src/components/choropleth.js
+++ b/src/components/choropleth.js
@@ -110,8 +110,7 @@ function ChoroplethMap({
         );
       svgLegend.attr('viewBox', `0 0 ${widthLegend} ${heightLegend}`);
 
-      /* Draw map */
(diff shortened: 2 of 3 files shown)


jeremyphilemon:
Looks good, I think we'll also have to edit the "Hover over a state/UT for more details" text too.

jeremyphilemon:
I think it'd be good to add the "double click/tap to view districts" below the state name.

Also do you think it would be possible to have said selected state row highlighted (hovered style) too?

Screen Shot 2020-04-20 at 4 35 26 AM

shuklaayush:
@jeremyphilemon Don't make me do this

jeremyphilemon:
@shuklaayush 😂

PR #1304, opened 2020-04-19
Category sorting and Google Tag tracker

samyakmehta:
**Description of PR**
This pull request includes two things:

- Bug fix : Sort Categories on Essentials page in alphabetical order
- Feature : Added Google Tag tracker for Google Analytics. The tracker used is configured to track usage of Essentials page

**Type of PR**

- [x] Bugfix
- [x] New feature

**Relevant Issues**  
Fixes #...

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
1. Category sorting in desktop
Sorting in Desktop

2. Category sorting in mobile
Sorting in Phone

3. Page View Tracking
Page View Tracking

4. Click Tracking
Click Tracking

5. Test data in Google Analytics
Google Analytics Screenshot


src/components/resources.js
```diff
@@ -165,7 +165,7 @@ function Resources(props) {
             });
           });
         });
-        return array.map((x, i) => (
+        return array.sort().map((x, i) => (
```
aasimafridi44, line 168: Seems looks to fine. But I checked in entire code of this file. And find we are using multiple foreach inside. That is running three loop inside. Which is total incorrect. As we know we are going to large level data. So we should consider small things that could be impact at performance of our application. 
See below example -

image

We could use loadash function to handle this scenario. That provide linq operation.

@samyakmehta @nikita-11

src/components/resources.js
```diff
@@ -187,7 +187,7 @@ function Resources(props) {
               if (array.indexOf(x) === -1) array.push(x);
             });
           });
-          return array.map((x, i) => (
+          return array.sort().map((x, i) => (
```
aasimafridi44, line 190: We should use meaning full variable name instead of 'x/i'. So that every developer can read and easy to understand the code. Code minify already doing sort our code during production build.

@nikita-11

aasimafridi44 (changes requested):

src/components/resources.js
nikita-11: @aasimafridi44 We can take these changes separately for sure. Please review the changes that current PR involves. Addition of GA Trackers.

nikita-11 (approved):

PR #1363, opened 2020-04-21
Improv: Add message for no results in Demographics

### LINKED ISSUE #789: Demography for dates when no new cases have been detected.
It seems like the website is not able to fetch the demographics data because data of the previously selected date is shown if there were no new cases on the selected date. 

"No New Case" label can be shown instead of the data of the previously selected date.

**Additional context**
I would like to work on it, if possible.
2f02c03a-3536-4614-8f2f-fc718d2f6533

shubhamjain2998:
**Description of PR**
In Demographics when a search is done resulting in no new cases, then currently previously fetched data is shown, added a message to prompt the user to modify search.
This issue had been opened 13 days ago, and still unresolved, that's why I have made a PR, sorry if it against the code of conduct, no bad intention.

**Type of PR**

- [x] Bugfix
- [ ] New feature

**Relevant Issues**  
Fixes #789 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
Screen Shot 2020-04-22 at 8 30 55 PM

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -2739,16 +2739,19 @@ footer {
     .react-date-picker__calendar-button {
       padding: 0 !important;
       margin: 0 !important;
-      margin-left: 0.5rem !important;
+      margin-left: 0.25rem !important;
       svg {
         height: 16px;
         width: 16px;
+        stroke-width: 3px;
       }
     }
 
-    .calendar-close {
-      width: 12px;
-      stroke-width: 3;
+    .react-date-picker__clear-button {
+      svg {
+        width: 16px;
+        stroke-width: 3px;
+      }
     }
 
     select {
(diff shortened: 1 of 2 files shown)


jeremyphilemon (changes requested):
Hey, thanks for the PR :D

Could you try using a different strategy like React.Suspense or something rather than introducing two state variables? Setting those two variables can cause unnecessary re-renders to the PatientDB. Let me know what you think!

shubhamjain2998:
@jeremyphilemon Ya We can use other strategies like suspense, but that will not benefit much I suppose, as 
- one state variable out of two is only for the initial rendering of the patient component, as when a user opens the Demographics page, raw data is fetched and some preprocessing is done, so that variable is for initial loading of the data only, we can also show a spinner there. so its only place suspense can be used.

- the other state variable checks if there is any successful result of the query entered by the user or not, in this case, we have to re-render the component as a user can modify his search which will result in a change in a select option of input field n number of times and we have to show the results accordingly. so here, we cant use suspense.

So that how I approached and understood this. :)

Your thoughts?

jeremyphilemon:
@shubhamjain2998 Fair enough, I shall approve :)

Feel free to open a PR adding the suspense part!

jeremyphilemon (approved):

PR #1366, opened 2020-04-21
Fix: Resolve bloodhound search on initial load or first remote query.

abdulghanitech:
**Description of PR**
The search bar on the homepage for searching cities, essentials, wasn't working fine on selecting something from predefined essentials from the given list.

Even after selecting an essential thing, the search didn’t happen, I've tried the enter key from keyboard or clicking the search icon nothing worked.

So, in order to fix this bug, I went through the [typeahead library](https://github.com/corejavascript/typeahead.js) which we're using in this project and got to know about the problem. And the problem was the remote data fetching is an asynchronous task and we were trying to handle it with a normal synchronous function. I just fixed it.

**Type of PR**

- [x] Bugfix
- [ ] New feature

**Relevant Issues**  
Fixes #...

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
ezgif com-optimize (1)

### DIFF
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -95,7 +95,7 @@ function Search(props) {
     };
 
     const essentialsAsync = (datums) => {
-      // to handle async remote call
+      // to handle async remote call on initial launch
       essentialsEngine.search(searchInput, essentialsSync);
     };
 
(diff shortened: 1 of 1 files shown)


jeremyphilemon:
@abdulghanitech THANK YOU SO MUCH FOR FIXING THIS! 

I can't tell you how much time I spent on trying to get this fixed. Great stuff!

Issue #675, opened 2020-04-05
District wise recovered, active, and deceased cases numbers


sam2964:
District wise report is not visible. No of total cases are visible but in that total no. of cases no. Of active,recovered,deceased figure is not  visible.

![Screenshot_2020-04-05-13-07-43-15](https://user-images.githubusercontent.com/56807608/78469395-78684b00-773e-11ea-871d-006659be3237.png)

sam2964:
#Type:bug

sam2964:
`enhacement`

sarkar-sayantan:
<img width="960" alt="covid1" src="https://user-images.githubusercontent.com/21034360/78866271-b5914d80-7a5c-11ea-922a-f1ab13f43651.PNG">
I was checking it. The reason behind this was - two different types of data is recieved. Look at the right side of the picture (highlighted).

sarkar-sayantan:
I think, I can fix it. Please assign it to me. @jeremyphilemon

sudevschiz:
This is because there is good data at district level for many of the recovered and deceased case. However, it's being worked on here to provide whatever data is available through : https://github.com/covid19india/api/issues/114

jeremyphilemon:
Closed by #1407

PR #1407, opened 2020-04-22
Feat: Add maps based on active, recovered, and deceased counts

### LINKED ISSUE #1388: Map showing percentage for active and deceased as well like confirmed
right now on the map when we hover we are seeing percentage of each state confirmed case
IMG_20200422_083546
so we are not able to see current active and deceased percentage as the overall percentage info is very helpful.
so i would like to have a drop down at the top the map saying confirmed, active and deceased so if user selects he wants to see active percentage the map would change to show current active for all the state and also the color would chnage to green and similarly we would have a color range for green as well. so in that case states like goa we can easily know it has fully recovered by seeing the state in green color.

this is in ref to the discussion what i did in issue #1345

**Additional context**

shuklaayush:
**Description of PR**
Refactored code to plot maps based on active, recovered and deceased counts. Map types are switched by clicking on the respective panels above the map. The colours represent absolute values and are not normalized in any way (population, total cases etc.).

**Type of PR**

- [ ] Bugfix
- [x] New feature

**Relevant Issues**  
Fixes #1388
Addresses #675

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
Peek 2020-04-23 17-25

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -429,11 +429,15 @@ h6 {
     width: 100%;
   }
 
-  .Minigraph,
-  .Level,
   .MapExplorer,
   .meta-secondary {
     width: calc(100% - 3rem);
+  }
+
+  .Minigraph {
+    .svg-parent {
+      width: calc(25%) !important;
+    }
   }
 
   .MapExplorer {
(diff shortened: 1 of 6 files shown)


keshank:
@shuklaayush Two suggestions.
1. If you can change the colour of borders for states when changing the map.
2. If you can add Tested to map as well

shuklaayush:
@keshank Thanks for the suggestions. This is WIP. Don't worry about 1 and other styling issues. As for 2, it'll probably be added later if/when the data is available at district level

gyanendrapatro:
I have changes which says about districtwise also

gyanendrapatro:
And also the color on map should be like current active 0 means the state is green that gives moreninfo to the users like its a green state

gyanendrapatro:
I have the changes if you want i can give a pr

Issue #1313, opened 2020-04-19
daily active cases gone negative in timeseries graph


nayakrahul:
**Describe the bug**
Suppose no. of active cases decreases from day X to day Y. That means there is no addition to no. of active cases for day Y. So my point is : should we show negative value or zero in the daily active cases timeseries graph for day Y.

**To Reproduce**

1. Go to the state page of Tamil Nadu
2. Scroll down to timeseries graphs for Tamil Nadu
3. Choose daily button
4. Notice the active graph
_Note : There are other states too with such scenario._

**Expected behavior**
Should we show negative values or zero ?

**Screenshots**
![Screenshot (25)_LI](https://user-images.githubusercontent.com/11428805/79696564-fa0dbc00-829a-11ea-8b4c-b983fef68db9.jpg)

**Addional Context
Normally the cumulative graph of confirmed, recovered and deceased cases will always be increasing except active cases. So such case will only happen in active cases.

nayakrahul:
You can assign it to me if you feel it to be relevant.

shuklaayush:
What's the point of showing zero instead of negative values for change in active cases?

Abhinav-Mani:
I agree with @shuklaayush. There is no point showing 0 in place of negative. 
I feel the daily active case has no significance, active cases should always be cumulative therefore the active case chart from daily should be removed .

nayakrahul:
I agree with @Abhinav-Mani . Daily active cases chart doesn't make any sense. 
Let me explain with an example
On day 1, person A, B and C are confirmed positive and on day 2, person D and E are confirmed positive but A dies and B recovers. So for day 1, daily active cases is 3. And for day 2, daily active cases is 0 (by formula: confirmed - recovered - deceased) but actually there are 2 active cases added in day 2.

shuklaayush:
I don't see what the confusion is. Active represents the number of people alive who have the virus right now. Thus, it captures the extent of the spread at this point in time. In your example, there were 3 such people on day 1, and on day 2, it remained at 3. Hence, the delta is 0. It makes perfect sense in my opinion.

Abhinav-Mani:
@shuklaayush  in current version many states are showing -ve active cases. So if a state X shows -10 active case for 19 April 2020. Does this have any significance ? Or what does it mean?

shuklaayush:
It means that the number of Active Covid-19 cases decreased by 10 on that day. This could be either because people recovered or because they died. 

[...]

PR #1426, opened 2020-04-22
Fixed number of active cases

vibvibgyor:
**Description of PR**

There was problem with the number of active cases. The formula to calculate the number of active cases is as follows:

> Active = Confimed - Recovered - Deceased

The problem with this is that if for any day the number of recovered + number of deceased is more than number of confimed then the number of active cases will be negative for any day, but this is logically wrong (_how can a number of active cases be negative for a day?_). The formula is updated to:

> Active = Max(0, Confimed - Recovered - Deceased)

This will also fix the problem with the time series 'Daily' graphs. This formula is also added for the total active cases. There might be a situation when this can also be negative.

**Type of PR**

- [x] Bugfix

**Relevant Issues**  
Fixes Number of daily active cases.

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Before:
Selection_037

After:
Selection_033

### DIFF
--- a/src/utils/commonfunctions.js
+++ b/src/utils/commonfunctions.js
@@ -66,10 +66,14 @@ export const preprocessTimeseries = (timeseries) => {
     dailyrecovered: +stat.dailyrecovered,
     dailydeceased: +stat.dailydeceased,
     // Active = Confimed - Recovered - Deceased
-    totalactive:
-      +stat.totalconfirmed - +stat.totalrecovered - +stat.totaldeceased,
-    dailyactive:
-      +stat.dailyconfirmed - +stat.dailyrecovered - +stat.dailydeceased,
+    totalactive: Math.max(
+      0,
+      +stat.totalconfirmed - +stat.totalrecovered - +stat.totaldeceased
+    ),
+    dailyactive: Math.max(
+      0,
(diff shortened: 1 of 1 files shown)


shuklaayush:
Please refer to the discussion in #1313 and reopen this if you think it's still valid.

vibvibgyor:
Aayush, I got the point that these are active cases are actually, "the change in the total Active ", but the other 2 charts are not "change in _x_ type of cases". 

So, should we rename the chart to "change in the _total_ Active ", as for daily basis, we can't account for total number of cases? Also as the other 2 charts are not depicting "change in cases, but rather" the actual cases" and using this chart as a a chart depicting "change" is actually _asymmetric_. 

One option is to rename the chart to "the change in active cases" (which is logically correct, but asymmetric from the given 2 charts) and other is to limit minimum value to 0.

What do you think?

shuklaayush:
@vibvibgyor The other three charts are also the "change in x type of cases". They just happen to be the same as the actual cases declared that day. It all boils down to the fact that the other three types of cases are always increasing. Thus, any change (delta) will be the same as the actual new cases. For active cases, there's a difference since they can decrease as well, and hence, the delta loses it's earlier meaning.

vibvibgyor:
@shuklaayush I got your point. Thanks for explaning.

PR #1431, opened 2020-04-23
adding maps for active and deceased

### LINKED ISSUE #1388: Map showing percentage for active and deceased as well like confirmed

gyanendrapatro:
**Description of PR**
Adding dropdown for easy access to see the active , confirmed and deceased percent both state and district level as well.
Also the count values on the districtwise also changes with type.
along with color changes.
**Type of PR**

- [x] New feature

**Relevant Issues**  
Fixes #1388 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
image

image

image

### DIFF
--- a/src/components/choropleth.js
+++ b/src/components/choropleth.js
@@ -18,6 +18,7 @@ function ChoroplethMap({
   selectedRegion,
   setSelectedRegion,
   isCountryLoaded,
+  type,
 }) {
   const choroplethMap = useRef(null);
   const choroplethLegend = useRef(null);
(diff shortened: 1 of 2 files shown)


gyanendrapatro:
helps on state page also
image

shuklaayush:
This has been added via #1407. Thanks for the PR though :)

Issue #959, opened 2020-04-11
Code splitting


iDhruv22:
**Is your feature request related to a problem? Please describe.**
No, it's not a problem but it would be good to have as our codebase is increasing.

**Describe the solution you'd like**
Split the code based on router and lazy load other routes than "Home" so its performance can be a boost. It'll be great for low internet speed users.

iDhruv22:
I would like to get assigned this to me.

bee-rickey:
Please speak to @jeremyphilemon before you pick this up for development.

iDhruv22:
##### Routing scheme - 
 - ***Demographics*** and ***Deep dive*** will have their own separate chunk.
 - The rest all will be in one main chunk.
##### Implementation - 
The react Suspense and lazy will be used to achieve this. As a fallback of Suspense, the spinner Icon will be considered

@jeremyphilemon can I start working on this?

+Edit
Added PR as well, let me know you have any feedback on top that.

PR #1475, opened 2020-04-24
Refactor: Lazy load routes and other conditional components

pavanmehta91:
**Description of PR**

A brief description of what changes your PR introduces. The better this description, the quicker we can review, exchange feedback, and merge!

**Relevant Issues**  
Fixes #...

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Add relevant screenshots here


jeremyphilemon:
Hey, thanks for the PR! Which issue is this PR based on?

pavanmehta91:
t's about lazy loading routes so that all of the code does not gets loaded together. It creates separate chunk files for js.
This may impact initial load times for people with slower network connectivity.
Please refer to this https://reactjs.org/docs/code-splitting.html
I just saw there was an issue for this before and no PR by the guy.
https://github.com/covid19india/covid19india-react/issues/959
It intends to solve the same problem. Though it's not a blocker but nice to have. It won't load libraries in the inner pages until not explicitly visited hence reducing the js file side for the first load

shuklaayush:
Looks like this was added with #1653. Thanks for the PR though! :)

PR #1478, opened 2020-04-24
Improv: Hide and reveal navbar on scroll for mobile devices

### LINKED ISSUE #1477: Sticky navbar
currently i found that our mobile users are facing difficulties in navigating between pages, they need to scroll up all the way to top, to navigate between pages. 

my solution is to make navbar fixed, show navbar when user scroll up on pages and hide when scroll down.

Binu42:
**Description of PR**
* this PR will make our user to navigate between pages easily in mobile devices without scrolling to top they can navigate easily.

**Relevant Issues**  
Fixes #1477 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
navbar-mob


src/components/navbar.js
```diff
@@ -18,11 +18,45 @@ const activeNavIcon = (path) => ({
 });
 
 function Navbar({pages, darkMode, setDarkMode}) {
+  if (window.innerWidth < 769) {
+    let prevScrollpos = window.pageYOffset;
+    window.onscroll = function () {
```
jeremyphilemon: I'm interested to know the implications of this scroll even handler, would it function only within the navbar component, and would it lead to any side effects in the scrolling of the home page (like scrolls getting cancelled and not being able to scroll smoothly)

Binu42: Hi, @jeremyphilemon 
this scroll event will apply in window width less than 769px and it will apply to all windows wherever navbar component is there.
``` javascript
if (window.innerWidth < 769) {
 ....    // this part of code in navbar can be moved to some other components also, but i found it suitable inside navbar component.
}
```

**regarding side effects** in the scrolling of the pages, i don't think it's going to effect. i have tested in this 7 browsers it's working smoothly as it should work.
WhatsApp Image 2020-04-24 at 11 09 42 PM

jeremyphilemon:
```
if (window.innerWidth < 769) {
}
```
This comparison fires for scroll of every pixel – even if the value is greater than 769, it still fires.

How about we use intersectionObserver API to detect if the user is away from the top part of the screen. and then activate the sticky?

Resource: https://github.com/streamich/react-use/blob/master/docs/useMeasure.md

Binu42: > i have changed few things in code to improve performance of scroll event.
* i moved `scroll event and if condition` code from navbar.js to App.js because in navbar `if condition` code was triggered each time when navbar get opened and when route changes.
* used setInterval to manipulate DOM inspite of manipulating DOM continously when scroll.

@jeremyphilemon 
Screenshot from 2020-04-25 12-16-27
> in the above fig, i have added few console.log line to  see when if condition and code inside setInterval is getting triggered. demo of that is show below.

i found that `if condition code` is running only once when App get mounted even for value greater than 769px and on resizing.

scroll

i can use the resources you shared, but i don't think it's going to make any big difference.
**Further your any advice and suggestions will be greatly appreciated.**

OstapKryvytskyy:
@jeremyphilemon, 
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver# is not supported on Safari on both Desktops and Mobile, so Apple users will be discarded if https://github.com/streamich/react-use/blob/master/docs/useMeasure.md is used (same as intersectionObserver API).
However the approach is perfect.
Thank you!

shuklaayush:
This probably requires too much work to merge. Create a new PR if it's still relevant

PR #1482, opened 2020-04-24
Feat: Highlight currently selected row in table

### LINKED ISSUE #1473: Visually highlight the currently selected row in table

Currently we have indicators like for map (border) and timeseries graph (dropdown) to show the selected values, however, that's now the case for tables. Highlighting it could largely help to visually navigate around with ease.

Highlight the rows that are selected. Essentially would have to use the same style as when row is in hovered state.

nayakrahul:
**Description of PR**
This PR highlights the currently selected row in table on change in map and timeseries charts. It has the same style as it is when hover.

**Relevant Issues**  
Fixes #1473

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
ezgif com-video-to-gif (3)

### DIFF
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -3,8 +3,11 @@ import Row from './row';
 import React, {useState, useEffect} from 'react';
 import {Link} from 'react-router-dom';
 
-const isEqual = () => {
-  return true;
+const isEqual = (prevProps, currProps) => {
+  return (
+    JSON.stringify(prevProps.rowHighlighted) ===
+    JSON.stringify(currProps.rowHighlighted)
+  );
 };
 
 function Table(props) {
@@ -270,6 +273,16 @@ function Table(props) {
                       state.state in districts
                         ? districts[state.state].districtData
                         : []
+                    }
+                    isHighlighted={
+                      !props.rowHighlighted.isDistrict &&
+                      props.rowHighlighted.statecode === state.statecode
+                    }
+                    highlightedDistrict={
+                      props.rowHighlighted.isDistrict &&
+                      props.rowHighlighted.statecode === state.statecode
+                        ? props.rowHighlighted.districtName
+                        : null
                     }
                     onHighlightState={props.onHighlightState}
                     onHighlightDistrict={props.onHighlightDistrict}
--- a/src/App.scss
+++ b/src/App.scss
@@ -1545,9 +1545,12 @@ table {
     tr {
       cursor: pointer;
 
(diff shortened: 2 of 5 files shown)


shuklaayush:
Nice! Can you add support for the district rows as well?

akashverma (commented):
nice work!

src/components/table.js
```diff
@@ -272,7 +272,16 @@ function Table(props) {
                         : []
                     }
                     isHighlighted={
-                      props.rowHighlighted === state.statecode ? true : false
+                      !props.rowHighlighted.isDistrict &&
+                      props.rowHighlighted.statecode === state.statecode
+                        ? true
+                        : false
```
shuklaayush: This is not required, right?

nayakrahul: Yeah

src/components/table.js
```diff
@@ -4,7 +4,7 @@ import React, {useState, useEffect} from 'react';
 import {Link} from 'react-router-dom';
 
 const isEqual = () => {
-  return true;
+  return false;
```
shuklaayush: Might be better to compare the props to check if they're unchanged, instead of always returning false. Otherwise, it would lead to extra renders.

jeremyphilemon (changes requested):
This looks good!

Rather than setting the state, can we pass the prop instead and check in order to limit the re-renders that occur? Essentially we would want a re-render to fire only for the said selected row if it's highlighted state (as a prop) changes.

nayakrahul:
If the Table component receives the same props as previous one, then the component will not be re-rendered. I am writing the comparison function for it as told by @shuklaayush .

jeremyphilemon:
@nayakrahul yes you're right, I missed that. 

I was gonna say instead of adding a new state variable in home to track the row highlighted, you can use `activeState` and `regionHighlighted` (for districts) instead.

nayakrahul:
ezgif com-video-to-gif (1)
@jeremyphilemon I printed a statement to the console when table component re-renders. You can see that the statement gets printed only when hovered state is changed in map (props changes).

jeremyphilemon (approved):

jeremyphilemon:
Well I was clearly wrong about the re-renders, but I was trying to make the case that a duplicate variable to track the highlighted need to be made because we're already tracking that in activeState and regionHighlighted. 

However, I'll take this anyway. Thanks :)

PR #1513, opened 2020-04-25
Feat: Add statistics section in state page.

### LINKED ISSUE #1456: Show more relevant statistics on state page
It will be good if a user will be able to see the statistics in percent if he goes to a specific state page, it depicts the performance of the state.

A Section showing all the statistics, after the mini graph, a very rough idea of what I suggest is
Screenshot from 2020-04-23 21-09-24
##### Yaa definitely with much better styling.
And moreover, we can also showcase doubling rate, cases per million, daily average cases or daily average recoveries

**Additional context**
I would like to work on this if it feels relevant.

shubhamjain2998:
**Description of PR**

Added relevant statistics like -
- Confirmed case percent of the total population (its shown correct to four decimal places as some states have a very low number of cases, rounding off to two decimal places will result in 0).
-  Active, recovered, and mortality rate correct to two decimal places.
- Cases per million and test per million.
- Positive Test rate as received in the API, and will show 'N/A' if data is not available.
- Doubling Rate, as calculated from the growth rate of the last 7 days.

To implement this section I added two files - 
- StateStats.json - containing a population of all states in JSON format, population numbers are referenced from here. [Link](https://en.wikipedia.org/wiki/List_of_states_and_union_territories_of_India_by_population)
- detailStats.js - A Separate Component to display the statistics section.

**Relevant Issues**  
Fixes #1456 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Screenshot from 2020-04-25 12-03-08
Screenshot from 2020-04-25 12-02-56


jeremyphilemon:
@shubhamjain2998 Thanks for the PR! :D

PR #1522, opened 2020-04-25
Feat: Navigation between state pages

### LINKED ISSUE #1372: Navigation between state pages
Navigation from one state page to a different state page without going back to the home page.

A drop down breadcrumb solution that lists links to all the other states would work well with the current design. 

A separate dropdown list would occupy too much space, especially on mobile.

**Additional context**
image

karankiri:
**Description of PR**

Added the navigation in breadcrumb so that user can travel from one state page to another state page without having to go to home page.

**Relevant Issues**  
Fixes #1372 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
image
image
image
image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -92,7 +92,9 @@ function App() {
                     <Route
                       exact
                       path={page.pageLink}
-                      component={page.view}
+                      render={({match}) => (
+                        <page.view key={match.params.stateCode || index} />
+                      )}
                       key={index}
                     />
                   );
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -92,7 +92,9 @@ function App() {
                     <Route
                       exact
                       path={page.pageLink}
-                      component={page.view}
+                      render={({match}) => (
+                        <page.view key={match.params.stateCode || index} />
```
jeremyphilemon, line 96: What does the `match.params.stateCode ||` do? 🤔

karankiri: `match.params.stateCode ||` will pass key as statecode for state page and index for other pages.
I'm passing in statecode from the url as key prop to forces the page re-render while choosing the different state from the breadcrumb dropdown.

jeremyphilemon: Oh, that's pretty cool actually! I'm wondering why `<Link>` doesn't do the re-render without this present though, any idea?

jeremyphilemon (approved):

Issue #1486, opened 2020-04-24
Website hangs when searched for essentials on smartphone


nprabhav:
**Describe the bug**
Site crashes when we search for essentials on Smartphones.

**To Reproduce**

1. Go to Menu
2. Click on Essentials
3. Click on Search
4. Type anything in Search Keyword
5. See error

**Expected behavior**
At least it should return 'No results found' when the search fails or it should return say top x items related to the search.

**Screenshots**

![issue1](https://user-images.githubusercontent.com/17107647/80206067-554bff80-8649-11ea-8bd1-98fc91de6737.jpg)

**Smartphone (please complete the following information):**

- Device: OnePlus6T
- OS: Android 10 OxygenOS 10.3.2
- Browser Chrome 80.0.3987.149
- Version 80.0.3987.149

raxhvl:
I can look into this.

itsplavika:
The same issue is happening on chrome Mac as well, the page becomes unresponsive
<img width="1680" alt="Screenshot 2020-04-25 at 6 18 20 PM" src="https://user-images.githubusercontent.com/29068720/80280366-4173cc80-8721-11ea-992c-0d65c1eb6c41.png">

nikita-11:
@ravindranrahul Are you working on this? This is a critical fix. Will need to add it soon.

nikita-11:
Else we planned to disable search bar functionality for a while. If possible, Let me know the updates on this.

raxhvl:
Yes, am working on this, meanwhile can we hotfix this by disabling search?

raxhvl:
It would be ideal if we merge #1532 and i can add the fixes on this page too.

jeremyphilemon:
@ravindranrahul 

>For issue #1486 solution might be the same which I have implemented.
Use the inbuilt useDebounce hook which is already defined in react-script - @chinmaydalvi

PR #1532, opened 2020-04-25
Fix: Debounce input events on search in home page

### LINKED ISSUE #1511: Search in home page freezes
**Describe the bug**
Search bar gets hang while finding something

**To Reproduce**

Go to 'Home Page'
Click on 'Search'
Start Typing
See error
Play around with the search, at some point it'll start to lag, and then hand completely.

**Expected behavior**
Search Bar should work properly without hanging.

chinmaydalvi:
**Description**

Fix the issue of Freezing of Search Bar while typing.
Used Debounce to fix the issue as well as to prevent unnecessary network calls & optimize the performance

**Relevant Issues**  
Fixes #1511 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

1. Typing Maharashtra
Typing Maharashtra (1)

2. Typing Madhya Pradesh
Typing Madhya Pradesh (1)

### DIFF
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -3,7 +3,7 @@ import {
   DISTRICTS_ARRAY,
   STATE_CODES_REVERSE,
 } from '../constants';
-import useDebounce from '../utils/hooks';
+import {useDebounce} from '../utils/hooks';
 
 import Bloodhound from 'corejs-typeahead';
 import React, {useState, useCallback, useRef, useEffect} from 'react';
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -3,9 +3,10 @@ import {
   DISTRICTS_ARRAY,
   STATE_CODES_REVERSE,
 } from '../constants';
+import useDebounce from '../utils/hooks';
 
 import Bloodhound from 'corejs-typeahead';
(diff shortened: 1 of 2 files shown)


chinmaydalvi:
@jeremyphilemon 
Please have a look as it can be High Priority.

jeremyphilemon:
@chinmaydalvi hey thank you, this looks good! I see that you've used a debounce hook. We're currently using `react-use` and it has a `useDebounce` [hook](https://github.com/streamich/react-use/blob/master/docs/useDebounce.md) already defined, would it be possible if you could use that if they're both performing the same function? :)

chinmaydalvi:
@jeremyphilemon Ok Let me try.

chinmaydalvi:
@jeremyphilemon 

Used `useDebounce` Hook  from  `react-use` &  working fine.

chinmaydalvi:
@ravindranrahul
For your issue #1486 solution might be the same which I have implemented.
Use the inbuilt  `useDebounce` hook which is already defined in `react-script`

jeremyphilemon:
Awesome, thanks :D

PR #1542, opened 2020-04-25
Feat: Added PWA Functionality

### LINKED ISSUE #1322: Convert it to progressive web app
The progressive web app will be very handy to use. People don't have to visit the website all the time they want to check status. They can easily install the website as an app in their android/ios devices. You can easily convert this website to a web app. ( progressive web app provide a banner for installing the website as an app in their phone )

Solution:
We need to just add a service-worker.js file in public/ directory and some additions in manifest.json will do it. it will not take that much time but it will create a great impact on users from my opinion.

Exact steps:

- create service-worker.js in public/ (Name should be exact same)
- Define an empty self.addEventListner("fetch" , () => ()) function like this for now.
- In maifest.json define orientation: "portrait primary" 
- In maifest.json add more icons ( 144x144, 512x512).

zephy20:
**Description of PR**

Added PWA functionality as requested by quite many people here. 

- Offline Support.

- Banner for "Add to Homescreen". 

**Relevant Issues**  
Fixes #1322 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**


jeremyphilemon:
Hey, so just letting you know that I had to revert the PWA functionality because updates to the website were not getting reflected. I read into this and it seems to be a problem with how PWAs work in general, and I think defining a network-first service worker would enable the updating functionality, but I still don't know too much about it.

We heavily rely on the webpage to be able to automatically update as we make changes to the codebase. 

Do you have an suggestions on how we can allow this and also get the benefit of it being a PWA?

zephy20:
Hey, sorry for the late reply.

For the problem you stated, we can use the combination of service worker methods `clients.claim()` and `skipWaiting()` which ensure that the service worker is updated as soon as there is a new deploy. You can read more about that here: https://davidwalsh.name/service-worker-claim

Let me know your thoughts on it. :)

PR #1555, opened 2020-04-26
Feat: Dynamic no. of rows with constant no. of columns

### LINKED ISSUE #1544: Dynamically set height of css grid when districts are expanded
**Describe the bug**
Untitled copy

The districts on expansion have a grid rule attached to it, however, due to a fixed height the width overflows. The workaround is using scroll to handle the overflow.

**To Reproduce**

1. Go to any Uttar Pradesh state page
2. Click on view all under top districts

**Expected behavior**
Ideally the height shouldn't be fixed, and it should expand dynamically keeping the no. of columns constant at three. The solution should eliminate the need to scroll.

nayakrahul:
**Description of PR**
The PR eliminates scrolling of all districts by setting the no. of rows dynamically. Calculated no. of rows by using total no. of districts to be shown.

**Relevant Issues**  
Fixes #1544 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
Screenshot (52)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -4040,14 +4040,15 @@ footer {
     display: flex;
     flex-direction: column;
     margin-right: 1rem;
-    width: 10rem;
 
     .button {
       background: $gray-light;
       color: $gray;
+      font-weight: 600;
       height: 2.5rem;
       justify-content: center;
       margin-top: .75rem;
+      text-transform: unset;
       width: 5rem;
 
       &:hover {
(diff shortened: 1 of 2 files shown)


kousiksatish (approved):
Good way of handling this :)

jeremyphilemon:
Thanks for the PR :D

PR #1568, opened 2020-04-26
Feat: code reduction using timeseriesexplorer.

### LINKED ISSUE #1557: Use TimeSeriesExplorer component in state pages.
Currently, home.js adds timeseries by adding the component `<TimeSeriesExplorer>` which contains the header and controls to operate the timeseries graphs. However, this is not the case in state pages, we're adding `<TimeSeries>` directly and rewriting the contents of `<TimeSeriesExplorer>` again over here.

Use `<TimeSeriesExplorer>` in state pages as well and handling the input by passing relevant props. You can use css to disable the select dropdown from rendering (similar strategy to how map explorer is re-used in state pages).

Open to more suggestions on abstractions if someone can do it elegantly, ideally would want to reduce code duplication.

Binu42:
**Description of PR**
* this PR will Reduce code in `state.js` using `timeseriesexplorer.js` inspite  of `timeseries.js` and writting html for select mode and tabs.

**Relevant Issues**  
Fixes #1557 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

### DIFF
--- a/src/components/state.js
+++ b/src/components/state.js
@@ -373,7 +373,7 @@ function State(props) {
               )}
 
               <div className="TimeSeriesExplorer">
-                <TimeSeriesExplorer stateTimeSeries={timeseries} />
+                <TimeSeriesExplorer timeseries={timeseries} />
               </div>
             </React.Fragment>
           )}
--- a/src/components/state.js
+++ b/src/components/state.js
@@ -4,7 +4,7 @@ import Footer from './footer';
 import Level from './level';
 import MapExplorer from './mapexplorer';
 import Minigraph from './minigraph';
-import TimeSeries from './timeseries';
+import TimeSeriesExplorer from './timeseriesexplorer';
 
 import {MAP_META, STATE_CODES} from '../constants';
 import {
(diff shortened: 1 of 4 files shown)


src/components/state.js
```diff
@@ -384,69 +373,7 @@ function State(props) {
...
-                  </div>
-                </div>
-
-                <TimeSeries
-                  timeseries={timeseries}
-                  type={graphOption}
-                  mode={timeseriesMode}
-                  logMode={timeseriesLogMode}
-                />
+                <TimeSeriesExplorer stateTimeSeries={timeseries} />
```
shuklaayush: Best if you can use the existing timeseries variable.

src/components/timeseriesexplorer.js
```diff
@@ -102,33 +103,39 @@ function TimeSeriesExplorer({
...
+        )}
       </div>
 
       <TimeSeries
-        timeseries={timeseries[activeStateCode]}
+        timeseries={
+          timeseries && activeStateCode
+            ? timeseries[activeStateCode]
+            : stateTimeSeries
+        }
```
shuklaayush: Might be better to instead pass only the state timeseries from both home.js and state.js. The complete timeseries data for all the states is not used anywhere in either TimeSeriesExplorer or TimeSeries. No need for the if conditions as well then.

src/components/timeseriesexplorer.js
```diff
@@ -102,33 +103,39 @@ function TimeSeriesExplorer({
...
         stateCode={activeStateCode}
```
shuklaayush: I know it's not related to this PR, but can you remove stateCode from here and from inside TimeSeries as well? I don't think it's used anywhere.

shuklaayush (changes requested):

shuklaayush (approved):
LGTM

Binu42:
@shuklaayush i forgot to remove this `<div className="TimeSeriesExplorer">` from state.js can you please remove it or i should raise another PR for this otherwise graphs will not occupy full space in mobiles.

shuklaayush:
@Binu42 No worries, just pushed a fix :)

Issue #1545, opened 2020-04-25
Create a build using webpack and do treeshaking to reduce the size of the build


faisal3389:
**Is your feature request related to a problem? Please describe.**
There are many places in the code where we are importing all the modules from the dependent library, even the ones which are not required. Eg:
import * as d3 from 'd3';
import * as Icon from 'react-feather';
import * as topojson from 'topojson';

here 'import *' imports all the modules which will bloat the size of the application.

**Describe the solution you'd like**
We could use a bundler such as rollup, which will help to treeshake and minify the js to reduce the build size

**Describe alternatives you've considered**
A more elaborate solution would be using webpack and building the js bundle through it.

ThePiyushGupta:
We can use webpack to build the js bundle through it . It is very easy but involves some manual verification . The documentation is descriptive enough https://webpack.js.org/guides/tree-shaking/ .

sreetamdas:
Doesn't create-react-app support tree shaking by default?
Also, @ThePiyushGupta are you looking at rewriting the app to use webpack? its using `react-scripts` right now, and I'd love to help with the rewrite :)

ThePiyushGupta:
I was trying earlier but it turned out to be very difficult and required some major changes. So now I am trying to use the rollup bundler.

ThePiyushGupta:
Yes u are correct create-react-app supports treeshaking by default.
Even after ejecting the project and rebuilding using webpack and making little config changes there were no changes.

Apart from it the parsing time is very less  as can be seen
![image](https://user-images.githubusercontent.com/16987573/80316706-f5a64d80-881c-11ea-95f5-3baa824b6a03.png)
We need to focus on other other areas to make the fronted snappy
Maybe caching some static files can be an option

jeremyphilemon:
@ThePiyushGupta @faisal3389 @sreetamdas +1 on the bundle size analysis, we're trying our best to minimize dependency on other packages, and trying to use vanilla js as much as we can.

With respect to using webpack, I'm assuming that would mean ejecting from `create-react-app`. [This](https://medium.com/curated-by-versett/dont-eject-your-create-react-app-b123c5247741) is a nice article addressing that, I feel besides reducing the bundle size, it could lead to increasing the complexity of the project configuration. But I'd like to hear from you if that's not the case.

We definitely need to look at different strategies too in order to have a faster frontend. #923 also addresses this.

cc @tannakartikey

ThePiyushGupta:
[...]

PR #1581, opened 2020-04-26
Refactor: Replace moment with date-fns

shuklaayush:
**Description of PR**
This PR tries to get rid of bugs arising due to time zone handling. All dates have been converted to JS date objects and references to moment have been removed.

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone


tannakartikey (approved):

tannakartikey:
@shuklaayush someone drew the attention on this:

> I don't think so that we can remove moment as moment is a dependency for charts.js

https://github.com/covid19india/covid19india-react/issues/1545#issuecomment-620110739

And they are right. https://www.chartjs.org/docs/latest/getting-started/installation.html#stand-alone-build

We may have to move forward according to the type of build of ChartJS we are using.

shuklaayush:
@tannakartikey Yeah, we're planning to remove ChartJS next

PR #1597, opened 2020-04-27
Default tab to Daily in Spread Trend

vibvibgyor:
**Description of PR**

This is from user's experience perspective. I realised that most of the time, I look for daily trends instead of cumulative trends (in Spread Trends). I surveyed around 30 people (with varied background, gender, age) and found that most (29/30) of the people are more interested in _Daily Trends_ than in _Cumulativee Trends_. Although the sample size is very small, but I have good confidence about more people are looking for the _Daily Trends_. This PR makes _Daily Trends_ as default tab.

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Before
Selection_041

After
Selection_042

### DIFF
--- a/src/components/timeseriesexplorer.js
+++ b/src/components/timeseriesexplorer.js
@@ -51,20 +51,20 @@ function TimeSeriesExplorer({
         <h1>Spread Trends</h1>
         <div className="tabs">
           <div
-            className={`tab ${graphOption === 1 ? 'focused' : ''}`}
-            onClick={() => {
-              setGraphOption(1);
-            }}
-          >
-            <h4>Cumulative</h4>
-          </div>
-          <div
             className={`tab ${graphOption === 2 ? 'focused' : ''}`}
             onClick={() => {
               setGraphOption(2);
             }}
           >
             <h4>Daily</h4>
+          </div>
+          <div
+            className={`tab ${graphOption === 1 ? 'focused' : ''}`}
+            onClick={() => {
+              setGraphOption(1);
+            }}
+          >
+            <h4>Cumulative</h4>
           </div>
         </div>
 


shuklaayush:
@vibvibgyor Hey, since the setting persists in local storage, is this really necessary?

vibvibgyor:
@shuklaayush Not able to get you?

shuklaayush:
@vibvibgyor We store this setting in the user's local storage. So when you open the website, it defaults to the setting which you had selected previously. See:
https://github.com/covid19india/covid19india-react/blob/1013f2f9bb9bebfab6aa0395cfcac0ead5495150/src/components/timeseriesexplorer.js#L15-L18

vibvibgyor:
Thanks for sharing this information @shuklaayush. I think the PR would be now relevant for the first time users, users who uses multiple devices, multiple browsers, incognito mode and may be after clearing the history/cache/site date.

shuklaayush:
@vibvibgyor Okay, I get your point :). But this is sort of a subjective issue. I, personally, would rather have it default to the cumulative charts when I open the website for the very first time, just to get an overview of the spread. Let's wait and see what other people think.

vibvibgyor:
Sure @shuklaayush :)

varshney-harshit:
agreed; I too look at the daily charts @shuklaayush @vibvibgyor 
can I make the required changes?

shubhamjain2998:
I also agree with @shuklaayush, I would prefer to see a cumulative chart if I am visiting this site for the first time, just to quickly get a rough idea about the trend through line graph.

vibvibgyor:
@shuklaayush https://www.worldometers.info/coronavirus/
They also replaced default tab from cummulative to daily.

shuklaayush:
@bee-rickey @sudevschiz @jeremyphilemon Your call.

shuklaayush:
Closing due to inactivity

PR #1623, opened 2020-04-28
Added Rollup and configured it to minify maps file

tannakartikey:
**Description of PR**
The JSON files inside `public/maps` are not minified as they are easy to develop.
But we should serve minified files on the production. The PR uses the non-minified version on development and the minified maps on the production.

Added a task `compress_maps` which calls Rollup.
This task is added in `build` task so that the maps are compressed on each deploy.

The original size of the folder is 1.7M and the size of the compressed folder is 916K.

**Relevant Issues**  

**Checklist**

- [ ] Properly formatted
- [ ] Tested on desktop
- [ ] Tested on phone

**Screenshots**

Add relevant screenshots here

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1820,10 +1820,38 @@
         "glob-to-regexp": "^0.3.0"
       }
     },
+    "@nodelib/fs.scandir": {
+      "version": "2.1.3",
+      "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.3.tgz",
+      "integrity": "sha512-eGmwYQn3gxo4r7jdQnkrrN6bY478C3P+a/y72IJukF8LjB6ZHeB3c+Ehacj3sYeSmUXGlnA67/PmbM9CVwL7Dw==",
+      "dev": true,
+      "requires": {
+        "@nodelib/fs.stat": "2.0.3",
+        "run-parallel": "^1.1.9"
+      },
+      "dependencies": {
+        "@nodelib/fs.stat": {
+          "version": "2.0.3",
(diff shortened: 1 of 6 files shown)


tannakartikey:
~~We need to modify the `build` command on ZEIT now when we merge this.~~

Fixed with: https://github.com/covid19india/covid19india-react/pull/1623/commits/ba34e3d628b340b0fefa04931f6c2b8bba3a6e03

shuklaayush:
@tannakartikey Can't we just always keep the map files as minified JSONs, and exclude them from prettier formatting? Won't have to add dependencies then

tannakartikey:
@shuklaayush actually it was minified first. Then we made it pretty.

It's hard to read the files and the properties when it is minified. Also, when the maps get updated, we can't tell what got updated.

Rollup is a dev dependency so it won't affect the performance of the app for the users.

jeremyphilemon:
@tannakartikey This build seems to be working on zeit, will it work on gh-pages as well? 🤔

tannakartikey:
@jeremyphilemon they both static servers so I don't think it will create trouble.

tannakartikey:
How are we deploying to gh-pages? As long as we have `maps_data` folder which is generated as a result of `npm run build` we should be fine.

jeremyphilemon:
@tannakartikey It still looks like the maps being downloaded on the production website aren't the minified versions? Am I seeing in the wrong place?

Screen Shot 2020-05-04 at 2 59 03 AM

tannakartikey:
@jeremyphilemon I am not sure how you are viewing this inside Source tab. I couldn't find the file there.

Anyways, following are the response we are getting from the server:
From `master`:
image


From this PR:
image


Notice that the whitespace is gone.

You will see the same if you run the task on local and then check the files inside `mini_maps`

jeremyphilemon (approved):

PR #357, opened 2020-03-29
Feature: Localization support

harshzalavadiya:
Revision of #132, Closes #21

## Changelog
- localization using `react-i18next`
- Compare to previous PR translations are not prebundled into code but instead will be loaded dynamically on switch

PR #1627, opened 2020-04-28
Feature/localization

tannakartikey:
**Description of PR**
Continuing from https://github.com/covid19india/covid19india-react/pull/432

**Relevant Issues**  
Related to #357

**Languages**

- [ ]  English
- [ ]  Hindi
- [ ]  Telugu
- [ ]  Kannada
- [ ]  Gujarati
- [ ]  Marathi
- [ ]  Tamil
- [ ]  Bangla
- [ ]  Odiya
- [ ]  Punjabi
- [ ]  Malayalam

**Checklist**

- [ ] Properly formatted
- [ ] Tested on desktop
- [ ] Tested on phone

**Screenshots**

Add relevant screenshots here


tannakartikey:
Steps to move forward:


#### Step 1:
Using [`useTranslation`](https://react.i18next.com/latest/usetranslation-hook) hook convert all the texts to translation tags(example below). It's okay if we don't have those keys in the locale files. This step is necessary to gather all the required keys and create locales. 
**Examples:**
https://github.com/covid19india/covid19india-react/blob/5c59c027918c00a0527c3c3c4f2fa307ba5d2ccd/src/components/row.js#L123
https://github.com/covid19india/covid19india-react/blob/5c59c027918c00a0527c3c3c4f2fa307ba5d2ccd/src/components/footer.js#L10


#### Step 2:
Next, we have to create locales out of these tags. One way is to use [i18next-parser](https://github.com/i18next/i18next-parser#cli) as shown below:
```
npx i18next 'src/**/*.js' -o public/locales/en.json
```
But the problem is that it is a static parser and can't fetch the dynamic data we receive from API. To overcome that we have to use [Runtime extraction](https://react.i18next.com/guides/extracting-translations#3-runtime-extraction). It reports the missing locales from the dynamic data with a post request to the given endpoint.

The server at [utility/missingLocaleServer.js](https://github.com/covid19india/covid19india-react/blob/5c59c027918c00a0527c3c3c4f2fa307ba5d2ccd/utility/missingLocaleServer.js) can catch these requests and writes them to a file called [`missing_locales.json`](https://github.com/covid19india/covid19india-react/blob/5c59c027918c00a0527c3c3c4f2fa307ba5d2ccd/missing_locales.json) at the root of the project.
It writes to that file instead of directly appending [`public/locales/en.json`](https://github.com/covid19india/covid19india-react/blob/5c59c027918c00a0527c3c3c4f2fa307ba5d2ccd/public/locales/en.json) because create-react-app keeps restarting with every change if we write inside `public/locales` directory.


#### Step 3:
Replace `public/locales/en.json` file with `missing_locales.json`. The `missing_locales.json` file uses the contents of `public/locales/en.json` so we will not lose the content of `en.json` file. This can be treated as our master locale file.


#### Step 4:
In many of the locale files, we have the state names and few other things translated in regional language. To use that we need to write a script which renames all the existing locales, except `en.json`, with `_old` suffix. For example, `hi.json` will become `hi_old.json`. Then the script we write will transverse through all the keys of `en.json` file and find its value from `hi_old.json` file. If the value is available the entry will be written inside `hi.json` else the key with the blank value will be added in the file.


#### Step 5:
Translation contributors can go ahead and edit the translation files and translate keys without values. 


#### Step 6:
When we update the content and there are sentences that have to be translated we get to know them from our `missing_locales.json` file. Any new code we write, we use translate tags from the beginning.

PR #1704, opened 2020-05-01
Fix: Make corrections to the map files

jeremyphilemon:


keshank:
The deployment has a bug with district Kutch in Gujarat. Which was not the case before.
Also the typo in search is broken as well.
Try Dahod in search. It shows up as "Dohad".
Try Dang it shows up as "The Dangs" in search

PR #1734, opened 2020-05-03
Add: Chart for deceased by age

vibvibgyor:
**Description of PR**

Added histogram for deceased by age.

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Selection_032

### DIFF
--- a/src/components/Charts/deceasedagechart.js
+++ b/src/components/Charts/deceasedagechart.js
@@ -1,0 +1,85 @@
+import {defaultOptions, xAxisDefaults, formatNumber} from './chart-defaults';
+
+import deepmerge from 'deepmerge';
+import React from 'react';
+import {Bar} from 'react-chartjs-2';
+
+function DeceasedAgeChart(props) {
+  const ages = Array(10).fill(0);
+  if (!props.data || props.data.length === 0) {
+    return <div></div>;
+  }
+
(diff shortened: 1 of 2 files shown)


bee-rickey:
I feel this would a misleading chart since the data set is small.

vibvibgyor:
The patient age chart has sample size of 2,344 patients out of 43,562 patients, which is around 5.3%. Now, when we compare sample size for deceased which is 44 out of 1,401 total deceased, this is around 3.3%. Even, the first chart doesn't give any significant results. I think, we should also remove the first chart. What do you think? @bee-rickey @shuklaayush

bee-rickey:
> The patient age chart has sample size of 2,344 patients out of 43,562 patients, which is around 5.3%. Now, when we compare sample size for deceased which is 44 out of 1,401 total deceased, this is around 3.3%. Even, the first chart doesn't give any significant results. I think, we should also remove the first chart. What do you think? @bee-rickey @shuklaayush

I think we should remove the other chart as well. +1 on that.

vibvibgyor:
Yeah. Please check #1785

shuklaayush:
Seems like this PR is obsolete now since deep-dive was deprecated. Closing

PR #1764, opened 2020-05-04
Fix: Memory leak due to setState in unmounted components & redundant Navbar re-render on each route

### LINKED ISSUE #1763: Memory leak due to setState in unmounted components
**Describe the bug**
Memory leak because of setState on unmounted components.

If an user goes to `/deepdive` (component mounted) from home but then immediately goes to `/faq` (`/deepdive` unmounted; `/faq` mounted) before `/deepdive` page could finish loading, the fetch requests in `/deepdive` fetch data in the background and then react tries to `setState` on (unmounted) `/deepdive`. This is causing memory leaks on all routes.

**To Reproduce**

1. In dev mode, go to homepage (http://localhost:3000/ if on default port) and then in Dev Tools -> Network -> set throttle to slow 3G. Don't disable cache.
2. Click on any other component/routes like Deep Dive/Essentials/FAQ and as soon as it starts loading, immediately go back to home page.
3. Keep checking network tab. If the chunk files for that component/route have loaded, repeat step 2.
4. Now once the JSON files have been fetched on slow 3G, console will show the memory leak error as it tries to setState on unmounted component after fetching has finished successfully.

**Expected behavior**
No `setState` on unmounted components.

**Screenshots**
mem-leak-all

- OS: Windows 10
- Browser: Chrome 
- Version: 81.0.4044.129 (Official Build) (64-bit)

abhishekakade:
**Description of PR**
- Added a check and cleanup function in `useEffect` to avoid setting state in unmounted components.
- Will fix memory leak caused due to setting state in unmounted components.
- Removed `Navbar` from `Suspense` in `App.js` to prevent redundant `Navbar` re-renders on each route.

**Relevant Issues**  
Closes #1763 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [ ] Tested on phone

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -112,15 +112,15 @@ function App() {
 
       <Router>
         <ScrollToTop />
-        <Suspense fallback={<div className="lazy"></div>}>
-          <Route
-            render={({location}) => (
-              <div className="Almighty-Router">
-                <Navbar
-                  pages={pages}
-                  darkMode={darkMode}
-                  setDarkMode={setDarkMode}
-                />
+        <Route
+          render={({location}) => (
+            <div className="Almighty-Router">
+              <Navbar
+                pages={pages}
+                darkMode={darkMode}
+                setDarkMode={setDarkMode}
+              />
+              <Suspense fallback={<div className="lazy"></div>}>
                 <Switch location={location}>
                   {pages.map((page, index) => {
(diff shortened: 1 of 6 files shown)


shuklaayush:
Closing due to inactivity

PR #1829, opened 2020-05-07
Service worker implementation using workbox

### LINKED ISSUE #1748: Add support for 'cache then network' strategy (PWA)
Yes, in order for us to be able to work in a PWA environment, we need a service worker that would update as we update the codebase so the patches made can be live.

Employ 'cache then network' policy. Here's the material for reference https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook#cache-then-network

Would like suggestions for a different approaches as well to support the above functionality in a PWA environment.

kousiksatish:
**Description of PR**
In this PR, I am trying to integrate workbox APIs to provide various caching strategies for files and enabling the app to work offline.

**Relevant Issues**  
Fixes #1748

**Checklist**
- [ ] Properly formatted
- [ ] Tested on desktop
- [ ] Tested on phone

PR not complete yet. Submitted for initial review

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1386,22 +1386,26 @@
     "@hapi/address": {
       "version": "2.1.4",
       "resolved": "https://registry.npmjs.org/@hapi/address/-/address-2.1.4.tgz",
-      "integrity": "sha512-QD1PhQk+s31P1ixsX0H0Suoupp3VMXzIVMSwobR3F3MSUO2YCV0B7xqLcUw/Bh8yuvd3LhpyqLQWTNcRmp6IdQ=="
+      "integrity": "sha512-QD1PhQk+s31P1ixsX0H0Suoupp3VMXzIVMSwobR3F3MSUO2YCV0B7xqLcUw/Bh8yuvd3LhpyqLQWTNcRmp6IdQ==",
+      "dev": true
     },
     "@hapi/bourne": {
       "version": "1.3.2",
       "resolved": "https://registry.npmjs.org/@hapi/bourne/-/bourne-1.3.2.tgz",
-      "integrity": "sha512-1dVNHT76Uu5N3eJNTYcvxee+jzX4Z9lfciqRRHCU27ihbUcYi+iSc2iml5Ke1LXe1SyJCLA0+14Jh4tXJgOppA=="
+      "integrity": "sha512-1dVNHT76Uu5N3eJNTYcvxee+jzX4Z9lfciqRRHCU27ihbUcYi+iSc2iml5Ke1LXe1SyJCLA0+14Jh4tXJgOppA==",
+      "dev": true
     },
     "@hapi/hoek": {
       "version": "8.5.1",
       "resolved": "https://registry.npmjs.org/@hapi/hoek/-/hoek-8.5.1.tgz",
-      "integrity": "sha512-yN7kbciD87WzLGc5539Tn0sApjyiGHAJgKvG9W8C7O+6c7qmoQMfVs0W4bX17eqz6C78QJqqFrtgdK5EWf6Qow=="
+      "integrity": "sha512-yN7kbciD87WzLGc5539Tn0sApjyiGHAJgKvG9W8C7O+6c7qmoQMfVs0W4bX17eqz6C78QJqqFrtgdK5EWf6Qow==",
+      "dev": true
     },
     "@hapi/joi": {
       "version": "15.1.1",
       "resolved": "https://registry.npmjs.org/@hapi/joi/-/joi-15.1.1.tgz",
       "integrity": "sha512-entf8ZMOK8sc+8YfeOlM8pCfg3b5+WZIKBfUaaJT8UsjAAPjartzxIYm3TIbjvA4u+u++KbcXD38k682nVHDAQ==",
+      "dev": true,
       "requires": {
         "@hapi/address": "2.x.x",
         "@hapi/bourne": "1.x.x",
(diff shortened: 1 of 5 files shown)


kousiksatish:
Currently, you can test this by reloading 2 times and then, going offline. Trying to reload again. It will start working offline. 
Below is the kind of strategy implemented
- precache - html, woff2
- staleWhileRevalidate - html, js, css
- networkFirst - json

Few pending tasks
- [ ] Need to figure out why we need to reload twice before it starts working fine. If it is possible to avoid this.
- [ ] Currently, other state pages are not cached unless visited at least once. Need to review fix.
- [ ] State level map json in home page is not cached unless visited atleast once. (Is this fine?)

@jeremyphilemon Review this and please feel free to suggest changes

shubhamjain2998:
@kousiksatish you need to reload the site 2 times because u are precaching only fonts and HTML and not other assets required to render homepage. When page loads first-time, service worker starts installing so it does not cache the necessary assets, and after refreshing the page , service worker cache all the assets. Here are screenshot depicting my analysis.
the cached data size for a fresh visit
Screenshot from 2020-05-08 01-43-21

the cached data size after refreshing page
Screenshot from 2020-05-08 01-43-33

kousiksatish:
@shubhamjain2998 I tried adding js and css to the precache initially. But what happened was, it was always fetching the files from precache. For js and css, we wanted to use staleWhileRevalidate so that latest changes are getting reflected when new version is updated. Is there any way to set both precache and staleWhileRevalidate, but give preference to staleWhileRevalidate

shubhamjain2998:
As far as i know precaching and stalewhilerevalidate are both different things, precaching will store the assests in cache and other will check if it become stale and needed to update so we need both.

Ya, and our webpage should always fetch the files from precache for offline support, and staleWhileRevalidate should update the precache stored in the browser.
I tried running the app with the same configuration as yours and I think its working fine when we are precaching all the assets of the home page. So Here's what I have done, correct me if I am wrong in understanding the whole scenario.
1. Included all the js and CSS files also for precaching.
2. Now When network is on, all assets are precached and ready for offline use.
3. Now when there is an update in the code and network is off then obviously site will show the old cached data, but when again network gets on **cache data is updated with the latest code** we are getting a console message for this as well
Screenshot from 2020-05-08 16-16-09
but now user has to reload the page to see the latest update. 
So we can do one thing that rather than showing a message in the console, we can make a separate component and show that message there as a notification so that the user will always get notified when there is an update.

kousiksatish:
@shubhamjain2998 I will verify what you had mentioned. If it is updating the stale copies for js, I will modify it.

Please have a look at this link: https://stackoverflow.com/questions/57780024/precaching-with-stale-while-revalidate-strategy-in-workbox

Do you think runtime caching will be better for our use case?

shubhamjain2998:
Yeah, I think runtime caching will be better for our use case as the cache will be continuously updated with the assets user is visiting, 
and by the way, with the above implementation using staleWhileRevalidate you are actually doing runtime caching already, like when you open any other page than the home in offline mode, it will not display anything as its cache is not present, now get back to online mode visit that page, its cache is added and now its also available in offline.
Refer this link for more clarity https://web.dev/runtime-caching-with-workbox/

kousiksatish:
@jeremyphilemon @shubhamjain2998 Have updated to precache js and css also. Please review

jeremyphilemon:
@kousiksatish @shubhamjain2998 Thanks so much for the PR and the discussions, really appreciate it! 

I tried to make a local build, however, I was greeted with this error after the main build is complete and proceeds to build the sw :/

```bash
> covid19india@1.0.0 build-sw /Users/jeremyphilemon/Clones/covid19india-react
> node ./src/sw-build.js

AssertionError [ERR_ASSERTION]: Unable to find a place to inject the manifest. Please ensure that your service worker file contains the following:/(precacheAndRoute\()\s*\[\s*\]\s*(\)|,)/
    at Object.<anonymous> (/Users/jeremyphilemon/Clones/covid19india-react/node_modules/workbox-build/build/entry-points/inject-manifest.js:86:5)
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/Users/jeremyphilemon/Clones/covid19india-react/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
    at _next (/Users/jeremyphilemon/Clones/covid19india-react/node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: null,
  expected: true,
  operator: '=='
}
```
Any idea what may be causing this?

shubhamjain2998:
its a weird error don't know why it is showing, as injection point for manifest file is defined in sw-template.js. make sure you are having all the files updated to the latest commit and try to build after reinstalling node modules, maybe that will do the trick.

kousiksatish:
@jeremyphilemon I also faced this when running building service worker file through workbox-cli. But didn't face it when running through workbox-build dependency.
Please clear build/service-worker.js file or the entire build directory before running build and let me know if it is solved.

jeremyphilemon:
@kousiksatish 

Had to revert :/

For some reason we're getting a generateSWString is not a function error during build. I'm pretty sure this wasn't the case before, and not sure what went wrong. I think I messed up somewhere during the squash, but I only edited the messages in the workbox. 

Also another suggestion, would it be possible to not cache api requests (json)? Without a "please refresh your page" it can be hard for people to realize that they would have to refresh the page in order to see the new stats. Thoughts?

---
I'm not sure if this library would help but check it out to see if it inspires a different approach https://github.com/zeit/swr

kousiksatish:
@jeremyphilemon I reviewed https://github.com/covid19india/covid19india-react/runs/676169371 
Does the CI script install dev dependencies? I have mentioned workbox-build as dev dependency.

I have given network first strategy for json. If network is online, I think the latest data will be fetched. If the network is offline, then it falls back to cached data. 

Are you referring to PWA. PWAs by default show the last loaded view and we may need to refresh to get latest. I can check out the best practice for this.

Will checkout zeit/swt and let you know.

jeremyphilemon:
@kousiksatish I would assume the ci does install dev dependencies since `rollup` is being used in the pipeline to compress the maps and it seems to be working fun during the build.

The thing with refresh is that, when I close the website and come back to visit it, it shows the cached version. It may not be immediately obvious for the person to refresh the page in that case since they've just arrive on the page (hoping that the latest data is fetched). If this isn't the case, then I think it's good!

Sure, I'll keep an eye on this PR.

PR #1842, opened 2020-05-09
Fix: NaN and Infinity in StateMeta Pages

vibvibgyor:
**Description of PR**

Some of the numbers in StateMeta pages were displayed as NaN and Infinity for some of the states. This has been fixed in this PR.

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Before (For Nagaland)
Selection_035

After (For Nagaland)
Selection_036

### DIFF
--- a/src/components/statemeta.js
+++ b/src/components/statemeta.js
@@ -22,14 +22,14 @@ function StateMeta({
   const sevenDayBeforeDate = format(lastSevenDaysData[0].date, 'dd MMM');
   const previousDayData = lastSevenDaysData[6].totalconfirmed;
   const previousDayDate = format(lastSevenDaysData[6].date, 'dd MMM');
-  var confirmedPerMillion = (confirmed / population) * 1000000;
-  var recoveryPercent = (recovered / confirmed) * 100;
-  var activePercent = (active / confirmed) * 100;
-  var deathPercent = (deaths / confirmed) * 100;
-  var testPerMillion = (lastTestObject?.totaltested / population) * 1000000;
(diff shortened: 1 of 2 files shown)


bee-rickey:
Lol @jeremyphilemon :P

vibvibgyor:
@jeremyphilemon @bee-rickey @shuklaayush Is something more required from my end here?

shuklaayush:
It seems like the files haven't been linted. The linter would've replaced references to 'var' with 'let'. Also, it would be great if you could reuse the formatNumber( ) method. It should ideally take care of NaN values.

vibvibgyor:
@shuklaayush Actually, I also thought to use this method, but there is no check for _Infinity_ in that function.

Shall I replace all the _var_ with _let_?

shuklaayush:
Just change the format function to check for isFinite() instead of isNaN().

> Shall I replace all the var with let?

Yes, but the linter should automatically do it for you. We only use let or const in the code.

vibvibgyor:
@shuklaayush Thanks, I have added the suggested changes in this commit.

vibvibgyor:
Please let me know, if something more is needed!

vibvibgyor:
@shuklaayush I have few more suggestions for the deep dive charts. Please close/merge this so that I could work further.

shuklaayush:
Hey @vibvibgyor, this seems to have been fixed while refactoring. Thanks for the PR though! :)

PR #1865, opened 2020-05-11
Correct the spell error of Kabirdham district in chhattisgarh in four…

RajatGupta8363:


shuklaayush:
Thanks, but please read [this](https://github.com/covid19india/covid19india-react/issues/662#issuecomment-609377552) before making a PR about district names

PR #1871, opened 2020-05-12
Odia localization file changed with correct translation

soumendrak:
**Description of PR**

Earlier, the translations were done with the help of Google Translate, which is recently launched on February end for Odia language and not yet absolutely correct.   
Modified few strings to capture correct translation of the terms. Thank you.

**Relevant Issues**  
N/A

**Checklist**
N/A
Testing not needed. These are only strings which got modified.

**Screenshots**
N/A


shuklaayush:
The localizations are now fetched from an API which pulls from a Google sheet that we maintain. Making changes here won't have any effect. Do reply here if you see major errors in the translations and would like to contribute

soumendrak:
I have made the changes jointly in the Google sheet also. No issues here. @shuklaayush Thanks.

PR #1872, opened 2020-05-12
Feat: Show nearby essentials. 

jeremyphilemon:
With reference to #1788


RONIZAM80:
Why so many issues in got hub

PR #1888, opened 2020-05-14
Use of ceil() function for Recovered/Deceased patients per 100 . 

### LINKED ISSUE #1885: Ceiling should be taken instead of round off.
**Describe the bug**
In state page, when we tell "x persons have passed away or recovered", we should consider ceiling not round off because if mortality rate is 1.4%, it means 2 persons have passed away not 1, we should consider 0.4% as 1 too. It makes sense when mortality rate is say 0.4% because according to current method, it will say "0 have passed away" which is obviously misleading. We are taking people into account. so a fraction should be considered ONE person regardless of whether it is less than 0.5 or more than. 

**To Reproduce**

1. Go to state page of Odisha
2. Scroll down to state metacards
3. See in mortality rate card that mortality rate is 0.49% but it is written "0 has passed away for 100 people" which is misleading.

**Expected behavior**
Ceiling should be taken not round off.

**Screenshots**
98000612_3002614863150405_3014041373536944128_n

KapoorAyush:
**Description of PR**

Use of ceil() instead of round() function for Recovered/Deceased patients per 100 patients. Reasons stated in the issue linked below. 

**Relevant Issues**  
Fixes https://github.com/covid19india/covid19india-react/issues/1885#issue-618124415

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Before
2020-05-14_LI

After
2020-05-14 (1)_LI

### DIFF
--- a/src/components/statemeta.js
+++ b/src/components/statemeta.js
@@ -105,8 +105,8 @@ function StateMeta({
           statistic={`${recoveryPercent.toFixed(2)}%`}
           formula={'(recovered / confirmed) * 100'}
           description={`For every 100 confirmed cases, 
-            ${Math.round(
-              recoveryPercent.toFixed(0)
+            ${Math.ceil(
+              recoveryPercent.toFixed(2)
             )} have recovered from the virus.`}
         />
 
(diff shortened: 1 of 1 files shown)


KapoorAyush:
@bee-rickey

bee-rickey:
@KapoorAyush We are currently discussing this internally. Adding a ceil might not be the right way since the deviation when the numbers are big is huge. Eg: if 4.44 is the number, and the state total tally is in 1000, then then technically you are saying 5 out of 100, or 50 out of 1000. While the actual count is 44. The issue magnifies with larger numbers. The better option is to change the way the sentence is formed so that it signifies approximations.

KapoorAyush:
@bee-rickey So should it say something like 4.44 out of 100?

shuklaayush:
Closing due to inactivity

PR #1916, opened 2020-05-17
Feature: Use Indian number system for yAxis all Graphs

### LINKED ISSUE #1787: Value in graph uses Indian numbering system, while axis uses international number system
**Describe the bug**
For example, in the "Tested" graph, the current value is shown as `11,07,233` while the Y-axis of the graph shows the numbers in millions like `1M`.

**To Reproduce**

1. Go to https://www.covid19india.org/
2. Scroll down to "Tested" graph on the bottom right.
3. Select "All States" to show data for all states. (Bug is there even without this but based on current numbers it shows only when "All States" is selected.

**Expected behavior**
Numbering system used in graph should be consistent with numbering system used in the axis.

**Screenshots**
Screenshot 2020-05-05 at 6 55 12 PM

- OS: OSX 10.15.4
- Browser: Chrome
- Version: 81.0.4044.129

**Additional context**
None

ayushya:
**Description of PR**

Data in the Graph uses Indian Numbering already, yAxis is using SI Numbering System, ie. 10Lakhs is represented as 1Million.

This PR Standardises the yAxis to use the Indian Numbering System.

**Relevant Issues**  
Fixes #1787 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Add relevant screenshots here
**Before**
Issue

**After**
After

### DIFF
--- a/src/components/timeseries.js
+++ b/src/components/timeseries.js
@@ -5,6 +5,7 @@ import {
   formatNumber,
   formatTimeseriesDate,
   formatTimeseriesTickX,
+  indianNumberSystemTickFormat,
 } from '../utils/commonfunctions';
 import {useResizeObserver} from '../utils/hooks';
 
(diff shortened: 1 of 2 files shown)


ayushya:
@shuklaayush 
Could you have a look at this PR, whenever you have some bandwidth?

shuklaayush:
Hi @ayushya, it might be better to keep the decimal digits as long as they're not trailing zeros. Otherwise, the y axis could have redundant entries. 

image

sudevschiz:
I think the SI system is the standard. It will help us in packaging the solution later and also in comparing the numbers with other countries. 

Since Indian school system teaches Lakhs and Millions equally, it's okay to retain this as is.

bee-rickey:
The standard used for graphs is the SI system. After an internal discussion, it feels like all the mods agree on this. Closing this for now.

PR #1917, opened 2020-05-17
Refactor switch route

harishmathur:
Refactoring App.js to simplify the use of Switch/Route and removing nested route as that does not seem to be getting used.


imbudhiraja (approved):

shuklaayush:
Closing due to inactivity

PR #1940, opened 2020-05-19
Added scroll-to-top button

RaghavK16:
**Description of PR**

- Updated the index page by adding **scroll-to-top button** for ease of back-scrolling.

- Added CSS, so button changes color when hovered over, or clicked.

- Web and Mobile compatible.


anukaal:
Can you add some screenshot of the updated page and the previous page shown

PR #2122, opened 2020-06-11
[FIX] move window to top when toggling language

arpitansu:
Move window to top when toggling the language (clicking the language toggle button)

Step 1: Scroll to the bottom or in the mid.
Step 2: Now click the Language Toggle button. The screen will not move to top creating confusion for the user.

**Checklist**

- [X] Compiles and passes lint tests
- [X] Properly formatted
- [X] Tested on desktop
- [X] Tested on phone

### DIFF
--- a/src/components/navbar.js
+++ b/src/components/navbar.js
@@ -32,6 +32,7 @@ function Navbar({
       <div
         className="navbar-left"
         onClick={() => {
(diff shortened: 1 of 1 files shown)


sparsh-99 (approved):

shuklaayush:
Fixed via #2220. Thanks for the PR though! :)

PR #2140, opened 2020-06-13
Fix description of average growth rate statistic

varunvora:
The formula suggested growth over the entire week but description mentioned it as daily average growth rate

### DIFF
--- a/src/components/statemeta.js
+++ b/src/components/statemeta.js
@@ -143,10 +143,9 @@ function StateMeta({stateCode, data, timeseries}) {
           description={
             growthRate > 0
               ? `In the last one week, the number of new infections has
-              grown by an average of ${formatNumber(
+              grown by ${formatNumber(
                 Math.round(growthRate / 7)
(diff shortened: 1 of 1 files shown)


shuklaayush:
Seems okay to me. If we change the description to this, then dividing by 7 doesn't make sense

varunvora:
@shuklaayush 
Dividing the weekly growth rate by 7 does not give daily average growth rate. Either the description or the formula should be fixed.
E.g. growth rate of 50% a week would mean 100 cases rose to 150.
A growth rate of (50/7)% i.e. 7.14% daily would mean cases rose to 100* (1.0714)^7 i.e. 162.
In the millions, these stats look misleading

shuklaayush:
@varunvora Ah, you're right. It's the simple vs compound interest thing. Let me fix it

shuklaayush:
Fixed via 60f2ac064c7f5cd7ce113697b0d6ab92877a5f1e

PR #2149, opened 2020-06-15
Slightly faster pageloads

slightlyoff:
**Description of PR**

Removes polyfills for `fetch()` (axios) and `IntersectionObserver` and delays loading of Google Analytics. In discussion with Jeremy it seems that losing compatibility with IE11 isn't a major issue.

The gains from delaying fetching of the GA scripts, and avoiding the gtm indirection, will be minimal for users on fast networks but should help those on slower links.

There's a lot of work remaining here to understand what the heck WebPack is doing for the default route and ensuring that we can preload the script that it's currently pulling in over a pretty janky stair-step, which you can see here at line 8:

https://www.webpagetest.org/result/200613_9S_fcc0999074dbca2982bc4e49dca95d21/1/details/#waterfall_view_step1

### DIFF
--- a/src/components/banner.js
+++ b/src/components/banner.js
@@ -1,6 +1,7 @@
-import axios from 'axios';
 import React, {useState, useEffect, useCallback} from 'react';
 // import {useTranslation} from 'react-i18next';
+
+const DATA_URL = 'https://api.covid19india.org/website_data.json';
 
 function Banner(props) {
   // const {t} = useTranslation();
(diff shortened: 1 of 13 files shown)


src/components/banner.js
```diff
@@ -1,22 +1,23 @@
-import axios from 'axios';
 import React, {useState, useEffect, useCallback} from 'react';
 // import {useTranslation} from 'react-i18next';
 
+const DATA_URL = 'https://api.covid19india.org/website_data.json';
```
jeremyphilemon, line 4: Just wanted to clarify in case you mistook this for the landing page's data, those endpoints would be:
- https://api.covid19india.org/v3/min/data.min.json
- https://api.covid19india.org/v3/min/timeseries.min.json

PR #2194, opened 2020-06-28
Update readme file with latest website image

swarnava:
**Before**:-

**Now**:-


shuklaayush:
Looks like this is outdated as well. Let's change it again later when we have a stable design :)

Issue #2182, opened 2020-06-23
Rate of positive test results


akumar90:
**Is your feature request related to a problem? Please describe.**
Have a clear understanding on whether the increased number of cases is because of increased testing Vs. increased spread. As per WHO guidelines, if a positive test rate is greater than 10%, the country is not doing enough testing.

**Describe the solution you'd like**
Understanding what the trend of rate of positive tests has been overtime.

**Describe alternatives you've considered**
N/A

**Additional context**
Having a graph similar to the active graph would be great. I am imagining the rate to be calculated as (number of positive results/number of tests)%

**Also working on creating something for rate of transmission of the disease. Will create a new feature ticket once I have requirements flushed out.

rtkg12:
I think this will be a good metric to add too and I'm interested in working on this.

agrahajigyasu:
This feature will be really helpful and will help people to judge which state is ahead in testing. Very interested to work on this feature

akumar90:
In addition to the positivity rate, creating a view for Rt (i.e. rate of transfer) would also be helpful.
I have taken the liberty of writing the code for these functionality (in Python though). Hopefully someone can re-use this code to be implemented as part of React.

Check **data_processing.ipynb**

[covid.zip](https://github.com/covid19india/covid19india-react/files/4880629/covid.zip)

PR #2205, opened 2020-07-03
Feat: Added tests positive percentage graph

rtkg12:
**Description of PR**

Added a new graph to show the positivity rate trend of tests which provides more insight into the current testing situation.

**Relevant Issues**  
#2182

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Screen Shot 2020-07-02 at 10 40 33 PM
Screen Shot 2020-07-02 at 10 40 46 PM

**Notes**

I have reused the same CSS styles for the confirmed statistic as I was not sure what color to use. I can change the color and name of the title if needed.

### DIFF
--- a/src/utils/commonFunctions.js
+++ b/src/utils/commonFunctions.js
@@ -114,6 +114,10 @@ export const getStatistic = (data, type, statistic, perMillion = false) => {
     const recovered = data?.[type]?.recovered || 0;
     const migrated = data?.[type]?.migrated || 0;
     count = confirmed - deceased - recovered - migrated;
+  } else if (statistic === 'percentPositive') {
+    const confirmed = data?.[type]?.confirmed || 0;
+    const tested = data?.[type]?.tested || 0;
+    count = tested === 0 ? 0 : confirmed/tested*100;
   } else {
     count = data?.[type]?.[statistic] || 0;
   }
--- a/src/App.scss
+++ b/src/App.scss
@@ -1507,6 +1507,10 @@ h6 {
       &.is-deceased {
         background: $gray-light;
       }
+
+      &.is-percentPositive {
+        background: $cherry-light;
+      }
     }
   }
 }
@@ -1589,6 +1593,17 @@ h6 {
 
     h4 {
       color: $gray-mid;
+    }
+  }
+
+  .is-percentPositive {
+    h1,
+    h5 {
+      color: $cherry;
+    }
+
+    h4 {
+      color: $cherry-mid;
     }
   }
 }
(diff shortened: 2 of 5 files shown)


src/utils/commonFunctions.js
```diff
@@ -114,6 +114,10 @@ export const getStatistic = (data, type, statistic, perMillion = false) => {
     const recovered = data?.[type]?.recovered || 0;
     const migrated = data?.[type]?.migrated || 0;
     count = confirmed - deceased - recovered - migrated;
+  } else if (statistic === 'percentPositive') {
+    const confirmed = data?.[type]?.confirmed || 0;
+    const tested = data?.[type]?.tested || 0;
+    count = tested === 0 ? 0 : confirmed/tested*100;
```
shuklaayush, line 120: Hey, the CSS we can sort out, but the primary problem is this calculation. Testing data and confirmed counts may include backlogs and hence, this won't give the actual positivity rate. The solution is to modify the API to include the number of positive samples out of the total samples tested. I think most states report this data. I'll see how it can be added to the API.

rtkg12: What do you mean by backlogs? Shouldn't that formula work fine since the confirmed cases and testing numbers are updated daily?

shuklaayush: Backlogs are when states reveal cases which were found to be positive at an earlier date but weren't reported in the state bulletin due to whatever reasons (mainly [bureaucratic](https://www.whatshot.in/bangalore/recorded-spike-bangalores-covid-19-accounting-error-c-23328)). 
> What do you mean by backlogs? Shouldn't that formula work fine since the confirmed cases and testing numbers are updated daily?

Unfortunately not. There's more nuance to it.

rtkg12: Oh okay, got it. I don't see how that should matter much since this is supposed to be an estimate anyway, but I can help out with the API integration if needed.

rtkg12: @shuklaayush Sorry to bother you, but do you have any updates on this? I think adding this graph for now and then updating it later if needed would suffice too since the cumulative graph should be a good enough estimate of the trend of the positivity rate.

sparsh-99 (approved):

rtkg12:
@sparsh-99 Thanks for approving. When will this be merged into master?

sparsh-99:
Its solely depends on the owner of the project.

shuklaayush:
Hey @rtkg12, sorry for the delay. I just added TPR to the table via #2282. The calculation is based on the number of positive samples, and not the people/confirmed value. We feel that the distinction is important. 

Adding TPR graph is straightforward now. Just have to edit the constants.js file and add a bit of CSS. The problem now is what to display in the TPR graph on the Daily tab.

Anyways, I'm closing this PR since it seems obsolete after the code changes. Thanks!

PR #2232, opened 2020-07-10
Feat: Add swipe support on Map Explorer

### LINKED ISSUE #2230: Swipe Support for toggling between confirmed, active, recovered & deceased status on home & state  page
On home page currently user can toggle between `confirmed, active, recovered & deceased` status using `colored dots`.
Map Image
It will be good if we can provide `long left & right swipe support ` on the map area  to toggle between these status.

On home page, if user swipes left/right on Status Map, it should change between prev/next status along with updating upper colored dots.

We can check for touch device & add support for listening swipe events & change current status based on it!

rahulbhooteshwar:
**Description of PR**

This change adds support of swipe gestures on Map Visualizer on home & state page to toggle between various statistics i.e. `Confirmed, Active, Recovered` etc.

**Relevant Issues**  
Fixes #2230

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

### DIFF
--- a/src/components/MapExplorer.js
+++ b/src/components/MapExplorer.js
@@ -172,13 +172,13 @@ function MapExplorer({
   });
 
   const swipeHandlers = useSwipeable({
-    onSwipedLeft: () => {
+    onSwipedRight: () => {
       const currentIndex = PRIMARY_STATISTICS.indexOf(mapStatistic);
       const toIndex =
         currentIndex > 0 ? currentIndex - 1 : PRIMARY_STATISTICS.length - 1;
       setMapStatistic(PRIMARY_STATISTICS[toIndex]);
     },
-    onSwipedRight: () => {
+    onSwipedLeft: () => {
       const currentIndex = PRIMARY_STATISTICS.indexOf(mapStatistic);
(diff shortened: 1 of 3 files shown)


src/components/MapExplorer.js
```diff
@@ -30,6 +30,7 @@ import React, {
 import {useTranslation} from 'react-i18next';
 import {useHistory} from 'react-router-dom';
 import {animated, useSpring} from 'react-spring';
+import {Swipeable} from 'react-swipeable';
```
shuklaayush: Since we use React hooks everywhere else in the code, do you mind replacing this with `useSwipeable`?

shuklaayush (changes requested):

src/components/MapExplorer.js
rahulbhooteshwar: Sure, I will make the changes. @shuklaayush can you mark the related issue #2230  as available, so that I can claim the same?

shuklaayush (approved):
LGTM. Thanks for the PR! :)

PR #2235, opened 2020-07-11
Fix: Sticky state meta

### LINKED ISSUE #2234: last updated and link for state page gets overflow when we try to see to testing stats
**Describe the bug**
last updated and link for state page gets overflow when we try to see to testing stats.

**To Reproduce**

1. Go to Home
2. Open district table of a state
3. Scroll right on table to see testing stats
4. Notice the error

**Expected behavior**
Position should be intact

**Screenshots**
Screenshot (90)

nayakrahul:
**Description of PR**

Fixed the issue by making last updated and state link sticky

**Relevant Issues**  
Fixes #2234 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Screenshot (93)
Screenshot (94)


shuklaayush:
Hi @nayakrahul, while it's a good jugaad, it doesn't actually fix the underlying problem. You can see that it's still there by adding more columns to the table. The primary issue is that the table-wrapper div is not expanding to the whole scrollable area.

nayakrahul:
> Hi @nayakrahul, while it's a good jugaad, it doesn't actually fix the underlying problem. You can see that it's still there by adding more columns to the table. The primary issue is that the table-wrapper div is not expanding to the whole scrollable area.

Yes @shuklaayush . It doesn't fix the problem properly. It's just a jugaad for current problem.

shuklaayush:
C'mon, you can do better :P

nayakrahul:
> C'mon, you can do better :P

sure. I am trying. :P

shuklaayush:
Fixed via 2cd279a898732db88a2e5ba11037c0e55d56d195. I sort of applied a different jugaad

PR #2242, opened 2020-07-14
Fix: Accessibility tab focus getting hidden

### LINKED ISSUE #2241: Accessibility Issue - Tab Focus is lost on Page
**Describe the bug**
Focus is lost on page while trying to access the site only by using keyboard. This is against accessibility rules. A user without mouse may never know where the actual focus is while trying to use the site by only using keyboard.

**To Reproduce**
Open the site. Keep clicking Tab button on keyboard.

**Expected behavior**
A clear indication should be given to the user of where the focus is.

shouvikbhuiyan1990:
**Description of PR**
Fixes the focus lost issue, if the user tries to access the site by only using keyboard.

**Relevant Issues**  
Fixes #2241 

**Checklist**

- [ ] Properly formatted
- [ ] Tested on desktop
- [ ] Tested on phone

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -938,7 +938,6 @@ h6 {
     color: $gray;
     font-size: 1rem;
     height: 2rem;
-    outline: none;
     padding: 1rem;
     padding-left: 3rem;
     width: calc(100% - 4rem);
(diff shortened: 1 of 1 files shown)


shuklaayush:
I'm pretty sure hiding the outline was an intentional design choice. @jeremyphilemon Thoughts?

shuklaayush:
Closing due to inactivity

PR #2255, opened 2020-07-18
Fix: swipe support for Map Visualizer

### LINKED ISSUE #2230: Swipe Support for toggling between confirmed, active, recovered & deceased status on home & state  page

rahulbhooteshwar:
Related to #2230
Swipe functionality got reversed during a review improvement.
Fixed it now.

**Description of PR**

This fix corrects swipe behavior on map visualizer, i.e. Left swipe go to previous & right swipe go to Next
**Relevant Issues**  
Fixes #2230 

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

### DIFF
--- a/src/components/MapExplorer.js
+++ b/src/components/MapExplorer.js
@@ -172,13 +172,13 @@ function MapExplorer({
   });
 
   const swipeHandlers = useSwipeable({
-    onSwipedRight: () => {
+    onSwipedLeft: () => {
       const currentIndex = PRIMARY_STATISTICS.indexOf(mapStatistic);
       const toIndex =
         currentIndex > 0 ? currentIndex - 1 : PRIMARY_STATISTICS.length - 1;
       setMapStatistic(PRIMARY_STATISTICS[toIndex]);
     },
-    onSwipedLeft: () => {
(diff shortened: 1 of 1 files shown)


rahulbhooteshwar:
@shuklaayush Can you please have a look?

shuklaayush:
Hey @rahulbhooteshwar, that was an intentional change from my side. Isn't that how swipe usually works on phones? For example, on the home screen, you swipe left and the screen moves right

rahulbhooteshwar:
@shuklaayush  I don't think so.. just matched current behavior with 2-3 apps, whatsapp , google photos to confirm. Our impl feels opposite.

rahulbhooteshwar:
@shuklaayush  My bad..got it now.. Thanks. Closing it.

PR #2256, opened 2020-07-19
Added Levitt metric

anirudhajith:
I've added a new tab (in addition to Cumulative and Daily) which shows a plot of the the [Levitt metric](https://thefederal.com/pandemic-1-mn-in-india/when-will-covid-19-pandemic-end-two-months-or-so-says-levitts-model/) vs time. This involved creating two new functions in src/utils/commonFunctions.js and addition of some code to src/components/Timeseries.js

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
Desktop view:
Added tab for Levitt metric (desktop view)")

Mobile view:
Added tab for Levitt metric (mobile view)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11373,9 +11373,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.15",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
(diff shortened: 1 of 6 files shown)


shuklaayush:
While I appreciate your work, we're trying to keep the metrics that we display as simple as we can so that it's easy to understand for common people. Otherwise, we get overwhelmed with queries regarding what those metrics represent. Hence, I'm closing this but thanks for the PR! :)

PR #2272, opened 2020-07-24
Use sqrt scale for choropleths (see screenshots)

aravindet:
**Description of PR**

For choropleth maps, change the scale from sequential to sequentialSqrt.

As the number of cases in a few districts are orders of magnitude above all the others, at an India-level district choropleths have a few dark spots in an ocean of white. It isn't possible to distinguish between the districts that have a statistic near the national median, and those districts for which it is close to zero.

See screenshots.

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**

Current:
Original

This PR (square root):
Sqrt

Alternative considered (log):

Log

Unfortunately I think this goes too far, minimizing the difference between hotspot districts and the others. It's very sensitive to the choice of the "bottom" of the scale (0 is not an option), and will need some work to set that to, say, the min or 1-percentile number from the data set. Also makes the legend at the bottom very crowded.

### DIFF
--- a/src/components/MapVisualizer.js
+++ b/src/components/MapVisualizer.js
@@ -22,7 +22,7 @@ import classnames from 'classnames';
 import {max} from 'd3-array';
 import {json} from 'd3-fetch';
 import {geoIdentity, geoPath} from 'd3-geo';
-import {scaleSqrt, scaleSequential} from 'd3-scale';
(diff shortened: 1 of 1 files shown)


shuklaayush:
Closing due to inactivity

Issue #2293, opened 2020-07-29
Add tooltip on footer icon


onshadeep:
**About adding tool tip on footer icon**
There are five icon on bottom footer. Can we put the tool tip on each icon?
![Screenshot_2020-07-29 Coronavirus in India Latest Map and Case Count](https://user-images.githubusercontent.com/43501083/88830436-94ab5e80-d1eb-11ea-9e64-051cfdb7b684.png)

KhanStan99:
A PR is raised for this issue. hope it gets merged.
https://github.com/covid19india/covid19india-react/pull/2263

jay1310:
Hey, can i work on it? Or it has been resolved?

venunalla7731:
According to PR #2263  footer icons also added tooltips. 
And issue resolves when PR Merged.

shreyanshi2228:
Can I work on this.

shreyanshi2228:
A tooltip is often used to specify extra information about something when the user moves the mouse pointer over an element. So, what is that extra information?

shuklaayush:
Closing due to inactivity

PR #2294, opened 2020-07-29
Tooltip issue resolved

jay1310:
**Description of PR**
It is regarding the issue #2293 for adding tooltip on footer icons.

**Fixes**
To solve the issue, I have added a dependency named react-tooltip and added properties to the anchor tag in the footer component.
Now, a tooltip is being shown for each footer icon as soon as mouse is hovered on the icon.

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [x] Tested on phone

**Screenshots**
Screenshot (709)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15111,6 +15111,22 @@
         }
       }
     },
+    "react-tooltip": {
+      "version": "4.2.7",
+      "resolved": "https://registry.npmjs.org/react-tooltip/-/react-tooltip-4.2.7.tgz",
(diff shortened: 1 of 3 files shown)


shuklaayush:
Hey, we dropped dependency on `react-tooltip` earlier due to bundle size and other issues (see #2074). Also, I don't think we need any fancy tooltips for this. HTML title tags should suffice. Thanks!

jay1310:
Okay. Sorry for that, but i had no idea that it had been dropped earlier.
Actually, I had tried it using title tags in the footer component elements, but it was not visually attractive, therefore, I used react-tooltip. Thank you for your response.

PR #2431, opened 2021-04-23
Make the node packages to be upgraded by itself

ProdeveloperProgrammer:


ravi-maithrey:
I don't think this works the way you think it does. If you use ">" there are many chances to develop weird edge cases. At least use ">=" Also sometime in the future, the project might go in a direction where ver 10 for some module might not be supported but version 11 is. With a modifier like >7, the package manager would make do if ver 10 was present which would cause errors. Give this a thought

arpit20adlakha:
This can create some serious issues like not taking care of  deprecations and breaking changes. Major versions have breaking changes.

PR #2486, opened 2021-05-26
Issue #2485 - Fix rounding Error

### LINKED ISSUE #2485: 0.3 rounded as 0 , makes statement wrong
**Screenshots**

image

**Describe the bug**

The above page shows `~0` which is marginally different from `0.3` as 0 sort of means none

**To Reproduce**

1. Go to 'https://www.covid19india.org/state/KL'
2. Scroll down to Case Fatality Ratio

**Expected behavior**
0.3 instead of 0 in the statement 

- OS: macOS Version 11.3.1 (Build 20E241)
- Browser :  Chrome
- Version : 90.0.4430.212&nbsp;(Official Build)&nbsp;(x86_64)

codelear:
**Description of PR**

The rounding off was presenting incorrect results when the percentage was less than 1. So I fixed this by checking if the percentage is between 0 and 1 and scaling the value by 10.

**Relevant Issues**  
Fixes #2485

**Checklist**

- [x] Properly formatted
- [x] Tested on desktop
- [ ] Tested on phone

**Screenshots**

image

### DIFF
--- a/src/components/StateMeta.js
+++ b/src/components/StateMeta.js
@@ -59,6 +59,24 @@ function StateMeta({stateCode, data, timeseries}) {
   const growthRate =
     (Math.pow(lastConfirmed / prevWeekConfirmed, 1 / 7) - 1) * 100;
 
+  const percentageDescription = (percentageValue, description) => {
+    let confirmedCasesText 
+
+    if (percentageValue > 0 && percentageValue < 1)
+    {
+      confirmedCasesText = "For every 1000 confirmed cases";
+      percentageValue = percentageValue * 10
+    }
+    else
+    {
+      confirmedCasesText = "For every 100 confirmed cases";
+    }
+
+    return `${t(confirmedCasesText)},  
+    ~${formatNumber(Math.round(percentageValue))} 
+    ${t(description)}`
+  }
+
   return (
     <>
       <div className="StateMeta population">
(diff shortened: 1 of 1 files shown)


codelear:
@sarathsp06 - Fixes Issue #2485

src/components/StateMeta.js
```diff
@@ -59,6 +59,24 @@ function StateMeta({stateCode, data, timeseries}) {
   const growthRate =
     (Math.pow(lastConfirmed / prevWeekConfirmed, 1 / 7) - 1) * 100;
 
+  const percentageDescription = (percentageValue, description) => {
+    let confirmedCasesText 
+
+    if (percentageValue > 0 && percentageValue < 1)
+    {
+      confirmedCasesText = "For every 1000 confirmed cases";
+      percentageValue = percentageValue * 10
```
sarathsp06, line 68: For semantics sake, i wouldn't call it `percentageValue`  anymore, preferably return it here or use a more appropriate name

src/components/StateMeta.js
```diff
@@ -59,6 +59,24 @@ function StateMeta({stateCode, data, timeseries}) {
   const growthRate =
     (Math.pow(lastConfirmed / prevWeekConfirmed, 1 / 7) - 1) * 100;
 
+  const percentageDescription = (percentageValue, description) => {
+    let confirmedCasesText 
+
+    if (percentageValue > 0 && percentageValue < 1)
```
sarathsp06, line 65: A small inconsistency here 
* Since `percentageDescription` is always called for percentageValue > 0 , no need to check for `percentageValue > 0`
* Its good to test `percentageValue > 0` but a case when percentage value <= 0 has to be handled then

shuklaayush:
Thanks @codelear for the PR and @sarathsp06 for the inputs. This should be fixed via 2a0fba85bcc8e608f0d2b3d9563513417aa8b6ab. 

I wasn't in favour of switching the text between per 100 and per 1000 based on the percentage value, so I modified the formatting logic to show decimals if the percentage is between 0 and 1. Closing this.

